PR #12, opened 2023-04-18
Push app settings

bjkeller:
Adds copy of viewer application settings from the template project


common/src/python/flywheel_adaptor/project_adaptor.py
```diff
@@ -160,3 +163,70 @@ def remove_gear_rule(self, *, rule: GearRule) -> None:
...
+        """
+        self.__fw.set_project_apps(project=self.__project, apps=apps)
+
+    def get_dataviews(self) -> List[DataView]:
+        """Returns the list of dataviews for the project.
+
+        Returns:
+          the dataviews in the enclosed project
+        """
+        return self.__fw.get_dataviews(self.__project)
```
dpark6060, line 189: This also ends up returning all dataview templates available to the project, even if they're not actually imported or used.  You may want to filter out only ACTIVE dataviews with the following check:

```
dataviews = self.__fw.get_dataviews(self.__project)
dataviews = [view for view in dataviews is view.parent != "site"]
```

dpark6060 (approved):
LGTM, I'll follow up on that modify dataview issue

PR #20, opened 2023-09-22
Gear pull scan metadata

bjkeller:
I think this is what needs to be done to make this a gear per my current understanding.

Most of the changes to code are formatting, but I did enable the FlywheelGearContext in run.py. I need to doublecheck how logging needs to be done b/c I have a feeling I was duplicating what is in the context class.

there is some refactoring that I'd like to do, but will leave until you are ready to merge


zstarkuw (commented):
So much whitespace change😵‍💫
Looks like a few minor changes, so if you were wanting to refactor before approval that's fine.

bjkeller:
OK, @zstarkuw refactored. 
Trying to get this so that we have code we can reuse in similar scenarios where we have CSVs in S3 that need to be partitioned and distributed to centers

PR #27, opened 2023-12-08
Generic parameters/Add RDS parameters

bjkeller:
This changes parameter methods so use a common generic method using Pydantic dictionary checking to eliminate duplicated code across methods.

Also, adds a method for pulling RDS parameters. Though may want to remove specialized methods at some point.

Note that using Pydantic in the way that is used here either requires Python 3.12 or above or using the typing-extensions package. Using typing-extensions here, but does cause some weirdness of types.
Also, it appears that VSCode Plyance doesn't handle some of the typing correctly.


hewanadun (approved):
Looks good.

PR #28, opened 2023-12-12
Pull identifiers from identifier-lookup-gear

bjkeller:
Adds a repository for identifiers. Currently, uses SQLAlchemy to connect to RDS identifier database.  Includes tests using local sqlite database.

### DIFF
--- a/common/src/python/identifiers/identifiers_repository.py
+++ b/common/src/python/identifiers/identifiers_repository.py
@@ -15,6 +15,9 @@ class IdentifierRepository:
     """Repository for Identifier records.
 
     Assumes the Identifier class is mapped to the identifier table.
+
+    Create repository within a resource block for a session to ensure that
+    session lifecycle is correct.
     """
 
     def __init__(self, session: Session) -> None:
--- a/common/src/python/identifiers/identifiers_repository.py
+++ b/common/src/python/identifiers/identifiers_repository.py
@@ -6,12 +6,9 @@ https://github.com/cosmicpython/code/tree/chapter_02_repository_exercise
 
 from typing import List, Optional, overload
 
-from identifiers.identifiers_tables import metadata
 from identifiers.model import Identifier
-from inputs.parameter_store import RDSParameters
-from sqlalchemy import create_engine
 from sqlalchemy.exc import NoResultFound
-from sqlalchemy.orm import Session, sessionmaker
+from sqlalchemy.orm import Session
 
 
 class IdentifierRepository:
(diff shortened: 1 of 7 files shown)


common/src/python/identifiers/identifiers_repository.py
```diff
@@ -0,0 +1,119 @@
...
+        database = 'identifier'
+        database_url = (f"mysql+mysqlconnector://{parameters['user']}:"
+                        f"{parameters['password']}@{parameters['host']}:"
+                        f"{port}/{database}")
+        engine = create_engine(url=database_url)
+        metadata.create_all(engine)
+        session = sessionmaker(bind=engine)()
+        return IdentifierRepository(session)
+
+    # def add(self, identifier: Identifier):
```
zstarkuw, line 45: extraneous, or TODO?

common/src/python/identifiers/identifiers_repository.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+class IdentifierRepository:
+    """Repository for Identifier records.
+
+    Assumes the Identifier class is mapped to the identifier table.
+    """
+
+    def __init__(self, session: Session) -> None:
+        self.__session = session
```
zstarkuw, line 24: I think the session need to be closed somewhere.

common/src/python/identifiers/identifiers_repository.py
```diff
@@ -0,0 +1,119 @@
...
+          List of all identifiers in the repository
+        """
+        if adc_id is None:
+            return self.__session.query(Identifier).all()
+
+        return self.__session.query(Identifier).filter_by(adc_id=adc_id).all()
+
+
+class NoMatchingIdentifier(Exception):
+    """Exception for case when identifier is matched."""
```
zstarkuw, line 119: is matched or is not matched?

common/src/python/identifiers/identifiers_repository.py
bjkeller: Good point. It seems like close() is more of a reset, but the recommendation is to use it anyway. 
I'll look at switching to using a resource block which makes sure close is called

common/src/python/identifiers/identifiers_repository.py
bjkeller: TODO, but removed

common/src/python/identifiers/identifiers_repository.py
bjkeller: I moved the "create_from" method out of the repository class and changed it to create a session. 

This can then be used in a resource block

```
identifiers_session = create_session(parameters)
with identifier_session as session:
    repository = IdentifierRespository(session)
```

which ensures that close is done at the end of the block.

common/src/python/identifiers/identifiers_repository.py
bjkeller: fixed

bjkeller: @zstarkuw  made changes described above

zstarkuw (approved):

PR #29, opened 2023-12-15
Pull error handling from identifier lookup

bjkeller:
Adds classes for capturing and outputting errors/alerts to CSV file
Based on current error output format from @dpark6060's file-validator.

Uses Pydantic model classes to define error data.
Modifies serialization so that when errors are dumped to dict, fields that are also models are serialized to JSON strings.

Includes a helper method for the error that occurs when a PTID isn't recognized when checking identifiers. (In that scenario, the center ID is fixed). Expect would want to write similar helpers for other scenarios.

ErrorWriter class is used to write errors to CSV, inserting Flywheel coordinates for the referenced input file.

### DIFF
--- a/common/src/python/outputs/BUILD
+++ b/common/src/python/outputs/BUILD
@@ -1,0 +1,1 @@
+python_sources(name="lib")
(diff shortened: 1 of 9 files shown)


hewanadun (approved):
Looks good.
For data quality check errors/alerts we need to add Error ID attribute (may be as an optional if we don't define it for other error types)

bjkeller:
I added error_id. Going to go ahead and merge

PR #30, opened 2023-12-20
Identifier lookup gear

bjkeller:
Note: this includes the code from #29 

Adds the identifier-lookup-gear: reads a CSV with ADCID and PTID on each row, looks up NACCID, if NACCID exists, outputs row to file, if ID doesn't exist outputs an error.


gear/identifer_lookup/src/python/identifer_app/run.py
```diff
@@ -0,0 +1,131 @@
...
+            sys.exit(1)
+
+        file_input = gear_context.get_input('input_file')
+        if not file_input:
+            log.error('Missing input file')
+            sys.exit(1)
+
+        file_id = file_input['object']['file_id']
+        adcid = get_adcid(proxy=proxy, file_id=file_id)
+        if not adcid:
```
zstarkuw, line 101: Can adcid be 0? We use that for the sample center. I think this would evaluate to True in that case

gear/identifer_lookup/src/python/identifer_app/run.py
```diff
@@ -0,0 +1,131 @@
...
+      adcid: the center ID
+    Returns:
+      the dictionary mapping from PTID to Identifier object
+    """
+    identifiers = {}
+    identifers_session = create_session(rds_parameters)
+    with identifers_session as session:
+        identifiers_repo = IdentifierRepository(session)
+        center_identifiers = identifiers_repo.list(adc_id=adcid)
+        if not center_identifiers:
```
zstarkuw, line 41: Can you explain this to me? 😅
I have to be missing something, but if center_identifiers returns nothing, is it setting identifiers from nothing?

gear/identifer_lookup/test/python/test_identifer_app.py
```diff
@@ -0,0 +1,184 @@
...
+    """Create data stream without expected column headers."""
+    data = [['dummy1', 'dummy2', 'dummy3'], [1, 1, 8], [1, 2, 99]]
+    stream = StringIO()
+    write_to_stream(data, stream)
+    yield stream
+
+
+@pytest.fixture(scope="function")
+def data_stream():
+    """Create data stream without header row."""
```
zstarkuw, line 56: Should this be 'with'?

gear/identifer_lookup/src/python/identifer_app/run.py
bjkeller: Yes. Replaced with explicit check for None

gear/identifer_lookup/test/python/test_identifer_app.py
bjkeller: yes

zstarkuw (approved):

PR #34, opened 2024-02-01
Fix gear rule copy

bjkeller:
Fixes an error in copying the fixed input of a gear rule so that file version could be mismatched.


bjkeller:
This is working and running on FW

PR #37, opened 2024-02-07
Write error metadata

bjkeller:
Switching error capture from writing CSV files to adding errors to file metadata.

### DIFF
--- a/common/src/python/outputs/errors.py
+++ b/common/src/python/outputs/errors.py
@@ -1,6 +1,6 @@
 """Utilities for writing errors to a CSV error file."""
 from abc import ABC, abstractmethod
-from typing import List, Literal, Optional, TextIO
+from typing import Any, Dict, List, Literal, Optional, TextIO
 
 from outputs.outputs import CSVWriter
 from pydantic import BaseModel, Field, field_serializer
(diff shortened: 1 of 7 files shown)


common/src/python/outputs/errors.py
```diff
@@ -25,46 +20,16 @@ class JSONLocation(BaseModel):
...
-        Args:
-          location: the location object
-        Returns:
-          JSON string representation of the location object
-        """
-        if not location:
-            return None
-
-        return location.model_dump_json()
-
```
hewanadun, line 51: don't we still need serialization for error_location?

hewanadun (approved):
Looks good. Please check the two comments.

common/src/python/outputs/errors.py
```diff
@@ -25,46 +20,16 @@ class JSONLocation(BaseModel):
 class FileError(BaseModel):
     """Represents an error that might be found in file during a step in a
     pipeline."""
-    error_type: ErrorType
+    error_type: Literal['alert', 'error'] = Field(serialization_alias='type')
+    error_code: str = Field(serialization_alias='code')
     error_id: Optional[str] = None
```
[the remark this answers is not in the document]
bjkeller: This was a change that Parker made on their side, which basically just flattens the structure

common/src/python/outputs/errors.py
bjkeller: No, since this will be stored in the metadata, it can be an object rather than a string

PR #39, opened 2024-02-14
Update retrieving parent containers

hewanadun:
gear_context.get_input_file_object_value() method returns an empty parents dict, change to retrieve parents from the file object instead.


bjkeller (approved):
Looks good

PR #48, opened 2024-03-15
Parameterize gearbot API key

bjkeller:
Changes the retrieval of the gearbot API key to allow for gearbots on different Flywheel instances.

Requires that a gear using a gearbot have a parameter in the manifest that is the prefix for `apikey` in the parameter store. 
So, if there is an API key at `/prod/flywheel/gearbot/apikey`, the prefix would be `/prod/flywheel/gearbot`.

The PR also updates all existing gearbot gears, and the manifest in the gear template.


hewanadun (approved):
Looks good.

PR #58, opened 2024-04-04
Add-flywheel-path

bjkeller:
This adds an adapted version of the [get_lookup_path](https://gitlab.com/flywheel-io/scientific-solutions/gears/file-validator/-/blob/main/fw_gear_file_validator/utils.py?ref_type=heads#L157) method from the fwreference class of the file-validator.

This computes the path to a container within flywheel


hewanadun (approved):
Gears invoked by a gear rule does not have access to group container, did a small fix in get_lookup_path() method to handle this.
I added timestamp to FileError class and added methods to set timestamp and flywheel_path in ErrorWriter classes.
Please check and merge.

PR #60, opened 2024-04-11
form-qc-checker CSV file processing

hewanadun:
Update form-qc-checker to use GearEngine
Add functionality for processing CSV inputs
Update error reporting to support CSV inputs

### DIFF
--- a/gear/form_qc_checker/src/python/form_qc_app/run.py
+++ b/gear/form_qc_checker/src/python/form_qc_app/run.py
@@ -2,10 +2,15 @@
 
 import logging
 import sys
+from typing import Optional
 
-from flywheel import Client
 from flywheel_gear_toolkit import GearToolkitContext
 from form_qc_app.main import run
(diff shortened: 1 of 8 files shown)


gear/form_qc_checker/src/python/form_qc_app/run.py
```diff
@@ -15,63 +20,102 @@
...
+
+def main():
+    """Load necessary environment variables, create Flywheel, S3 connections,
+    invoke QC app."""
 
-        run(fw_client=fw_client,
-            s3_client=s3_client,
-            gear_context=gear_context,
-            redcap_connection=redcap_connection)
+    try:
```
bjkeller, line 113: GearEngine.run handles the GearExecutionError so the exception handling is not necessary. I did that to reduce the duplicated code.

bjkeller (approved):
Looks good. Just a minor detail on using the GearEngine

PR #61, opened 2024-04-16
Single project templating

bjkeller:
Changes push template app to push one template project to matching projects within the groups for centers.

1. uses centers in nacc/metadata to identify center groups.
2. applies template project identified in manifest config


gear/push_template/src/python/template_app/run.py
```diff
@@ -49,19 +53,34 @@ def create(
...
 
         return TemplatingVisitor(
             admin_id=context.config.get("admin_group", "nacc"),
             client=client,
+            template_group=group_id,
+            template_label=template_label,
             new_only=context.config.get("new_only", False))
 
     def run(self, context: GearToolkitContext) -> None:
-        template_map = get_template_projects(group=self.admin_group(
```
echeng06, line 59: Can remove the import for this, but also curious if there's a need for this specific method/file now that you've moved the pattern finding to the `TemplateProject` class. It's fine as it's "common" code but afaik it isn't being used anywhere else.

bjkeller: Would be fine to remove it

echeng06 (approved):

echeng06 (approved):

PR #63, opened 2024-04-19
Fix user roles and permissions

bjkeller:
Makes assignment of user roles and permissions more consistent.

1. Ensures that metadata and center-portal projects have admin user permissions set either when created, or reloaded.
2. Ensures that admin user roles are set for center group projects when created or updated


bjkeller:
I'm going to merge this without review.

PR #84, opened 2024-08-08
Add redcap-project-creation gear

hewanadun:
Automatically creates REDCap projects for a study using super API token and XML project templates for modules.

This gear expects a YAML input file in the format [gear/redcap_project_creation/data/test-data.yaml ](https://github.com/naccdata/flywheel-gear-extensions/blob/redcap-project-creation/gear/redcap_project_creation/data/test-data.yaml)
XML templates are retrieved from NACC admin project in Flywheel.

For newly created REDCap projects, 

- Project level API token is saved in AWS parameter store using the REDCap project PID. 
- REDcap PID and module info saved in Flywheel metadata project for the respective study/center
- Same metadata info is written to a output YAML file in the destination project

### DIFF
--- a/common/src/python/centers/center_group.py
+++ b/common/src/python/centers/center_group.py
@@ -580,7 +580,7 @@ class REDCapFormProject(BaseModel):
 
     redcap_pid: int
     label: str
-    report_id: Optional[int]
+    report_id: Optional[int] = None
 
 
 class FormIngestProjectMetadata(IngestProjectMetadata):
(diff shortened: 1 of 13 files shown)


common/src/python/centers/center_group.py
```diff
@@ -709,3 +709,29 @@ class REDCapProjectInput(BaseModel):
...
     projects: List[REDCapFormProject]
+
+
+class REDCapModule(BaseModel):
+    """Information required to create a REDCap project for a module."""
+    model_config = ConfigDict(populate_by_name=True,
+                              alias_generator=AliasGenerator(alias=kebab_case))
+    label: str
+    title: str
+    template: Optional[str] = None
```
bjkeller, line 720: This is a reference of some sort, correct? Maybe add some detail to the class document block.

bjkeller (dismissed):
this looks good. Just a note on comment for "template" in redcap project object

common/src/python/centers/center_group.py
hewanadun: It is the XML template name (if different from module label), updated the docstring.

PR #87, opened 2024-08-29
Generalize guid pattern

bjkeller:
deja vu?


bjkeller:
the branch had not been deleted

PR #89, opened 2024-09-04
Optional forms validation

hewanadun:
List of optional forms vary by module/packet. Use [optional_forms.json](https://github.com/naccdata/nacc-draft-schemas/blob/main/rule-definitions/optional_forms.json) file to define optional forms.
For each optional form there are two definition files depending on whether the form is submitted or not

- Example rule definition [Form A2 - submitted](https://github.com/naccdata/nacc-draft-schemas/blob/main/rule-definitions/UDSV4/I/rules/a2_rules.json)
- Example rule definition [Form A2 - skipped](https://github.com/naccdata/nacc-draft-schemas/blob/main/rule-definitions/UDSV4/I/rules/a2_rules_optional.json)

The gear will load the correct definition file depending on the value of the **mode** variable for the respective form


bjkeller:
@hewanadun Where does the *mode* come from? I don't see it in the example files

bjkeller (changes requested):
This looks good, but can you make a pass at documentation in docs/form-qc-checker/index.md and explain how this works?   doesn't have to be exhaustive

PR #90, opened 2024-09-05
Change-enrollment-suffix

bjkeller:
This replaces occurrences of PTENRLv1 and ptenrlv1, which was mostly in tests and documentation.


hewanadun:
@bjkeller I made the following changes, please check and merge.
Add a function to the InputFileWrapper to get the module name from file suffix.
Add a gear config to the provisioning gear for the expected module name, default to 'enrollv1'
In the provisioning gear, checks whether the file suffix matches the expected module name.

PR #94, opened 2024-09-17
Add-registry-to-user-mgmt

bjkeller:
Changes user management gear to use the  comanage registry.

1. Creates registry record for new users in directory.
2. Creates new Flywheel users for claimed registry record.
3. Sends notification emails


hewanadun (approved):
Looks good!

hewanadun:
Noticed that study_id is set to 'adrc' when setting user permissions in Flywheel, hope this is intentional and managing other studies is a future task?

bjkeller:
In the Authorization class? That's a default. There are tests demonstrating using a different study id

PR #107, opened 2024-10-06
Improve-enrollment-errors

bjkeller:
Adds validation error handling to the creation of enrollment and transfer records.

Also adds code that allows dates to be parsed with more than one format.

### DIFF
--- a/common/src/python/dates/form_dates.py
+++ b/common/src/python/dates/form_dates.py
@@ -3,7 +3,7 @@
 from datetime import datetime
 from typing import List
 
-DATE_FORMATS = ['%m/%d/%Y', '%Y-%m-%d']
+DATE_FORMATS = ['%m/%d/%Y', '%m-%d-%Y', '%Y/%m/%d', '%Y-%m-%d']
 
 
 def parse_date(*, date_string: str, formats: List[str]) -> datetime:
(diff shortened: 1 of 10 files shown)


common/src/python/dates/form_dates.py
```diff
@@ -0,0 +1,32 @@
+"""Utilities to handle dates."""
+
+from datetime import datetime
+from typing import List
+
+DATE_FORMATS = ['%m/%d/%Y', '%Y-%m-%d']
```
hewanadun, line 6: QC rules accept either - or /, so all 4 formats '%m/%d/%Y', '%m-%d-%Y', '%Y/%m/%d', '%Y-%m-%d' are accepted

common/test/python/enrollment/test_serialization.py
```diff
@@ -1,9 +1,23 @@
...
 from pydantic import ValidationError
 
 
+@pytest.fixture
+def bad_date_row():
+    yield {
+        'adcid': 0,
+        'ptid': "123456",
+        'naccid': "000000",
+        'frmdate_enrl': "10/06/2024",
```
hewanadun, line 16: Isn't this a valid date format?

hewanadun (dismissed):
Please see the comments below.

common/src/python/dates/form_dates.py
bjkeller: changed

common/test/python/enrollment/test_serialization.py
bjkeller: yes. removed the slashes

PR #110, opened 2024-10-11
Add CHANGELOGs for all gears

echeng06:
Most of the CHANGELOGs are bare-bones but I tried to add detail where I could - ideally going forward these should be filled out alongside any updates to the corresponding gear.


bjkeller:
@echeng06 Can you add guidance for keeping them up-to-date in the developer document?

PR #115, opened 2024-10-25
Update build to force linux platform

echeng06:
Even if this is not our final solution I'd like to get this merged in just to unblock gear development in the interim. 

Main changes:
* Adds Go to the VSCode devcontainer
* Updates root BUILD file to specify default platforms for pex and docker builds
    * As part of this, adds `linux_x86_py311.json` (copied directly from the data platform demos), which is required for the pex binary build


danjonpeterson (approved):
Looks great! Especially good to see the changelog.

PR #118, opened 2024-10-31
Update csv-to-json-transformer gear

hewanadun:
- Creates a JSON file for each record in input CSV file
- Normalizes the visit date to YYYY-MM-DD format
- Creates Flywheel hierarchy (subject/session/acquisition) if it does not exist
- Check whether the record is a duplicate
- If not duplicate, upload/update file
- Update file info metadata and modality (have to do this here, running metadata extraction gear after upload makes things complicated for QC checks coordination)
- For each participant, creates a visits pending for QC file and upload it to Flywheel subject (QC Coordinator gear is triggered on this file)

**TODO**
- Currently only supports UDS, FTLD, and LBD hierarchy in Flywheel. Need to change to use a template file to support other modules. 
- Obtain required fields for each module from a template file
- Remove columns depending on the LBD version submitted
- May need to rename columns?


gear/csv_to_json_transformer/src/python/csv_app/transformer.py
```diff
@@ -0,0 +1,238 @@
...
+                filename=filename,
+                file_id=file_id,
+                visitdate=input_record[FieldNames.DATE_COLUMN])
+        else:
+            participant_visits = ParticipantVisits.create_from_visit_data(
+                filename=filename, file_id=file_id, input_record=input_record)
+            visit_mapping = {'subject': subject, 'visits': participant_visits}
+            self.__pending_visits[subject_lbl] = visit_mapping
+
+    def transform_record(self, input_record: Dict[str, Any],
```
bjkeller, line 119: I would like to separate the transformation from the file upload. Or at a minimum change the name of this method to imply we are creating the JSON file.

Also, we should factor out common code with EnrollmentSubject.upload_enrollment and create a method in SubjectAdaptor

gear/csv_to_json_transformer/src/python/csv_app/run.py
```diff
@@ -0,0 +1,95 @@
...
+            context.metadata.add_qc_result(self.__file_input.file_input,
+                                           name='validation',
+                                           state='PASS' if success else 'FAIL',
+                                           data=error_writer.errors())
+
+            context.metadata.add_file_tags(self.__file_input.file_input,
+                                           tags=context.manifest.get(
+                                               'name',
+                                               'csv-to-json-transformer'))
+            if sys_errors:
```
bjkeller, line 82: it seems weird to me to have the gear add qc results and file tags and *then* exit with an error. If it is OK to keep executing after the error occurs, just log it. Otherwise, raise an exception and make sure it gets caught and transformed into a GearExecutionError.

bjkeller (changes requested):
I would like to make sure we work out the commonalities b/w this and the identifier provisioning gear, which also creates JSON files attached to an acquisition under a subject.

We need to make common code for saving a file to a subject. See the note in the transformer code.

I think one of the key differences is that identifier provisioning is accumulating the objects to be saved and saving them outside of the code building the objects. If this were to follow that pattern, the transformer would take a dictionary and return the transformed dictionary and the objects would be accumulated and saved after.  I'm OK with this not following that pattern, but the word "transformer" needs to be changed.

I also made a note about the handling of system errors. I think to a certain extent this is caused by intermingling the transformation of the dict and saving to the file.

gear/csv_to_json_transformer/src/python/csv_app/run.py
hewanadun: Is it ok to raise an exception and stop processing the reset of the visits if a system error occurred while processing a visit?
If we just log and proceed, these errors will go unnoticed unless someone explicitly checks the log.

bjkeller: What is the chance that one file has a system error on save and the rest don't also fail?

An alternative would be to log it and send an email to nacc_dev@uw.edu

hewanadun:
@bjkeller I refactored the code and separated transformations from  upload. Still working on adding C2/C2T and LBD specific transformations, but please check whether the general structure looks fine to you. Thanks!

bjkeller (approved):
This looks good.

PR #121, opened 2024-11-18
Fix study mapping

bjkeller:
Fixes an error in how the study mapping was built, and refactors the code to pull pipeline construction out of the center group class.

### DIFF
--- a/common/test/python/projects/test_study.py
+++ b/common/test/python/projects/test_study.py
@@ -10,12 +10,12 @@ class DummyVisitor(StudyVisitor):
     """Visitor for testing apply methods."""
 
     def __init__(self) -> None:
-        self.center_name: Optional[str] = None
+        self.center_id: Optional[str] = None
         self.project_name: Optional[str] = None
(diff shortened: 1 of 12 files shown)


common/test/python/projects/test_study.py
```diff
@@ -65,7 +65,7 @@ def test_apply(self):
                         center_id="dummy",
                         adcid=1)
         center.apply(visitor)
-        assert visitor.center_name == "Dummy Center"
+        assert visitor.center_id == "Dummy Center"
```
echeng06, line 68: This test fails - `common/src/python/projects/study.py` needs to update the `Study.apply` method to say `visitor.visit_center(self.name)` instead of `visitor.visit_center(self)`

echeng06 (changes requested):

common/test/python/projects/test_study.py
bjkeller: Fixed. The error was in the Center.apply method

PR #128, opened 2024-12-05
Center Management: Unify CenterInfo logic

echeng06:
Consolidates `projects.study.Center` and `centers.naccgroup.CenterInfo` to `centers.center_info.CenterInfo` and updates the Center Management gear (updated to 1.0.3) along with any other effected common code to utilize it.

* New module `centers.center_info` was made since it makes more organizational sense and also leaving it in either of the previous modules causes some circular dependency issues
* `CenterMapInfo` remains unchanged but was also moved to `centers.center_info`
* A few additional tests also added for both

Also adds a small script to set up a gear for local run. 

Test to create a new dummy center using this version can be seen in the sandbox: https://naccdata.flywheel.io/#/jobs/6752536a2372f6b5625a9bec


common/src/python/centers/center_info.py
```diff
@@ -0,0 +1,76 @@
...
+                f"tags={self.tags}")
+
+    def __eq__(self, __o: object) -> bool:
+        if not isinstance(__o, CenterInfo):
+            return False
+        # compare everything except tags
+        return (self.adcid == __o.adcid and self.group == __o.group
+                and self.name == __o.name and self.active == __o.active)
+
+    @validator("tags")
```
bjkeller, line 44: this appears to be deprecated, use @field_validator instead

common/src/python/centers/center_info.py
```diff
@@ -0,0 +1,76 @@
...
+    adcid: int
+    name: str
+    group: str = Field(validation_alias=AliasChoices('center_id',
+                                                     'center-id',
+                                                     'group'))
+    active: Optional[bool] = Field(validation_alias=AliasChoices('active',
+                                                                 'is-active',
+                                                                 'is_active'),
+                                   default=True)
+    tags: Optional[Tuple[str, ...]] = ()
```
bjkeller, line 28: I think we are only using the tags for the ADCID. I think we should drop them

gear/center_management/src/docker/BUILD
```diff
@@ -6,5 +6,5 @@ docker_image(
     dependencies=[
         ":manifest", "gear/center_management/src/python/center_app:bin"
     ],
-    image_tags=["1.0.1", "latest"],
+    image_tags=["1.0.3", "latest"],
```
bjkeller, line 9: maybe go to 1.1.0

bjkeller (changes requested):
This looks good.

I would like to remove tags but maybe that should be a different PR

common/src/python/centers/center_info.py
bjkeller: Yes, let's punt on this. We should probably review the use of tags

bjkeller (approved):

PR #130, opened 2024-12-10
Revise transformer

bjkeller:
These changes are meant to generalize CSV to JSON transformation beyond the UDS/Form scenario.
Specific changes for this are:

* Make transformation schema an optional input file that is used to define transformations applied to the data. At this point, these are mostly dropping fields from the input records. Also, allows for having no transformations.
* Create a generalized JSON uploader to upload the record after transformed. And, adds a config parameter defining templates for labeling session and acquisition and naming the JSON file.


hewanadun (approved):
- It was little hard to track all the changes but the general flow seems to be fine.
- It seems some required fields checks got lost while trying to generalize the gear but since it is separated into two gears again we can add those back.
- Let's merge these changes first and I'll add those back along with any necessary pre-processing checks.

PR #131, opened 2024-12-10
Legacy identifier transfer

brendanrsmith:
This pull request introduces a new gear for transferring legacy NACCIDs to enrollment records in Flywheel. This includes handling validation errors and ensuring no duplicate subjects are created. This gear runs from a center's `ingest-enrollment` project and creates enrollment entries for each NACCID entry in the Identifiers database that does not currently exist in Flywheel for that center's ADCID. 

Outstanding questions: 

- [x] Do we need to check for a given NACCID at all other centers before creating an enrollment? 
- [x] Where to log errors? Currently using the FileError type but we aren't actually processing a file. Should we create a new error type? 
- [x] Do we need to handle enrollment record creation in batches?


bjkeller (commented):
Answers to your questions:
- You can assume the database is the source of truth on NACCIDs so you don't have to look anywhere else. And, at the moment, a NACCID only corresponds to a participant at one center
- Use logging to stdout. The file errors are meant for submitted data from centers.
- No you don't have to batch the record creation

common/src/python/outputs/errors.py
```diff
@@ -244,3 +244,22 @@ def errors(self) -> List[Dict[str, Any]]:
           List of serialized FileError objects
         """
         return self.__errors
+
+
+def legacy_naccid_error(value: str,
```
bjkeller, line 249: we should just use standard logging for these errors, so I don't think you need this

gear/legacy_identifier_transfer/src/python/legacy_identifier_transfer_app/main.py
```diff
@@ -0,0 +1,212 @@
...
+from enrollment.enrollment_project import EnrollmentProject
+from enrollment.enrollment_transfer import EnrollmentRecord
+from gear_execution.gear_execution import GearExecutionError
+from identifiers.model import CenterIdentifiers, IdentifierObject
+from outputs.errors import ErrorWriter, ListErrorWriter, identifier_error, legacy_naccid_error, unexpected_value_error
+
+log = logging.getLogger(__name__)
+
+
+class LegacyEnrollmentBatch:
```
bjkeller, line 18: I don't think it is necessary to batch these if it is easier to do otherwise. Batching in the identifier-provisioning gear is due to the input being a file with a batch of form records.

gear/legacy_identifier_transfer/src/python/legacy_identifier_transfer_app/main.py
```diff
@@ -0,0 +1,212 @@
...
+
+    def __iter__(self):
+        return iter(self.__records.values())
+
+
+def process_legacy_identifiers(
+    identifiers: Mapping[str, IdentifierObject],
+    enrollment_date: datetime,  # Added parameter for enrollment date
+    enrollment_project: EnrollmentProject,
+    error_writer: ErrorWriter,
```
bjkeller, line 46: you can drop the error writer and just use standard logging

gear/legacy_identifier_transfer/src/python/legacy_identifier_transfer_app/main.py
```diff
@@ -0,0 +1,212 @@
...
+            log.warning('Skipping record with missing NACCID: %s', record)
+
+    def __len__(self) -> int:
+        return len(self.__records)
+
+    def __iter__(self):
+        return iter(self.__records.values())
+
+
+def process_legacy_identifiers(
```
bjkeller, line 42: My general take is this function is too long and complex.
It might get better if you drop the error writer, but I think it would still be too long.
Maybe factor out some other functions that do task on a single identifier.

bjkeller:
@brendanrsmith I suggested some changes, but it looks good in general.

But, can you also run the following at the top level of the repo

```
pants fix ::
pants lint gear/legacy_identifier_transfer::
```

and `pants check gear/legacy_identifier_transfer::`

and fix anything that comes up?

brendanrsmith:
@bjkeller Changes made. One question that remains is if we want to handle the `enrollment_date` for the legacy naccids. Currently we are just setting that date to now.

hewanadun (commented):
Looks good overall, some minor comments
- In LegacyIdentifierTransferVisitor `__get_adcid()` rename the parameter `project_id` to `group_id`
- In main.py `run()`
    - adcid is unused parameter
    - Change the error message on `not success` to "Error(s) occurred while processing/importing legacy identifiers" [Saying no changes made is not accurate as some identifiers can get added]
- In `process_record_collection()`, I would change the error on `find_subject()` to a warning since this is kind of expected if we re-run this gear
- Merge the latest changes form main branch - enrollment classes are outdated

For the question regarding enrollment date, I'm not sure whether it is recorded in the identifiers database (not recorded in NACCMASTER). Another option is to set it to the date of the initial visit packet for that adcid, ptid. We can discuss.

PR #133, opened 2024-12-13
REDCap Error Checks Importer

echeng06:
New gear for importing error check CSVs into REDCap. 

Some uncertainty/questions I had:

* Are there other checks worth doing to "validate" the error check CSVs?
* I didn't love adding the `skip_sniffer` parameter to `read_csv`, but even after trying to standardize the CSVs to UTF-8 the sniffer still fails here. I've been having trouble with the Sniffer in other gears as well, so not sure if we should maybe remove it altogether or add another kind of `read_csv` method that explicitly doesn't use it?
* `read_csv` requires an ErrorWriter which requires a container ID and flywheel ID, which afaik isn't needed here so I end up just passing a dummy container/flywheel ID. This also feels like a hack, are there any cleaner alternatives?


echeng06:
Converting to a draft again since it needs to be cleaned up a bit after some last-minute fixes.

bjkeller:
> Some uncertainty/questions I had:
> 
> * I didn't love adding the `skip_sniffer` parameter to `read_csv`, but even after trying to standardize the CSVs to UTF-8 the sniffer still fails here. I've been having trouble with the Sniffer in other gears as well, so not sure if we should maybe remove it altogether or add another kind of `read_csv` method that explicitly doesn't use it?

We removed the use of sniffer from the reader already, but just a general point, csv.Sniffer is using heuristics that expect CSVs to follow consistent conventions and seem to fall apart when the values in a column may have characters that look like delimiters. For instance, having a value that has a list with commas in it.

The reader class was meant to be used in a case where we wanted to set rigid limits on how users provide data. If you are outside of this case, it would be OK to use different code.

> * `read_csv` requires an ErrorWriter which requires a container ID and flywheel ID, which afaik isn't needed here so I end up just passing a dummy container/flywheel ID. This also feels like a hack, are there any cleaner alternatives?

The error writer is setup to create error metadata that is meant to be read by end users. If the gear isn't processing user input for something derived from it, you should just use logging.

echeng06:
@bjkeller ah okay, yeah a lot of those in the error check CSVs (commas in lists) so that makes sense!

For `read_csv` I could update the method to write to an error writer if provided, and just log information otherwise? Does that work?

bjkeller:
I think adding a condition to each of the calls error_writer.write would obfuscate the code more than it already is.

Or, and this does seem like the long-way-around, even to me.  But may be more flexible in the long run. 
1. Rename ErrorWriter to something else (e.g., UserErrorWriter), 
2. make a new ErrorWriter as an abstract base class w/o the container nonsense, and 
3. then make a base class of the more general base class that just uses the logger for write.

bjkeller:
Also, can you fix the typing errors? `pants check ::`

echeng06:
@bjkeller I like the "long-winded" way, I agree it'll be more flexible in the long run. I can add that + fix the type checks.

common/src/python/outputs/errors.py
```diff
@@ -220,16 +262,15 @@ def write(self, error: FileError, set_timestamp: bool = True) -> None:
...
+        self.prepare_error(error, set_timestamp)
         self.__writer.write(error.model_dump(by_alias=True))
 
 
-class ListErrorWriter(ErrorWriter):
+class ListErrorWriter(UserErrorWriter):
     """Collects FileErrors to file metadata."""
 
     def __init__(self, container_id: str, fw_path: str) -> None:
+        """Initializer."""
```
bjkeller, line 273: Either go without a comment or say more

gear/redcap_error_checks_import/src/python/redcap_error_checks_import_app/visitor.py
```diff
@@ -0,0 +1,135 @@
...
+    def visit_header(self, header: List[str]) -> bool:
+        """Adds the header, and asserts all required fields are present.
+
+        Args:
+          header: list of header names
+        Returns:
+          True if the header has all required fields, False otherwise
+        """
+        error = None
+        for h in self.REQUIRED_HEADERS:
```
bjkeller, line 63: Not that I expect a change, but we should converge on a strategy for checking these. I have some code elsewhere that uses the set.subset operation to check this

bjkeller: an example: https://github.com/naccdata/flywheel-gear-extensions/blob/0f4f8c6ae3bf03bbb5f31054af00fb1012594ad3/gear/identifier_provisioning/src/python/identifier_provisioning_app/main.py#L110

bjkeller (approved):
None of my comments are breaking, so I'm going to go ahead and approve, but please take a look

common/src/python/outputs/errors.py
```diff
@@ -181,29 +183,69 @@ def previous_visit_failed_error(prev_visit: str) -> FileError:
...
+    def write(self, error: FileError, set_timestamp: bool = True) -> None:
+        """Writes the error to the logger.
+
+        Args:
+          error: the file error object
+          set_timestamp: if True, assign the writer timestamp to the error
+        """
+        if set_timestamp:
+            self.set_timestamp(error)
+        self.__log.error(error.model_dump(by_alias=True))
```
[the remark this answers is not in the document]
echeng06: I updated it to do a JSON dump for now, but we could always do a cleaner print (maybe remove null values? format it into a message?)

PR #135, opened 2024-12-13
Remove csv.Sniffer

echeng06:
Removes the csv.Sniffer. Tested in my other two gears (CSV Error Checks Importer + CSV splitter) and seemed fine, but the other gears that utilize `read_csv` should also be tested. 

The only functionality we can't really recreate is determining if something has a header, which I think can be ambiguous to begin with. I think it's fair to require that to be something that `visit_header` does; aside from the dummy tests, every implementation of `CSVVisitor` and `visit_header` expects _something_ from the headers. 

Another thing I noticed is that almost every implementation of CSVVisitor takes in an error writer as an initialization variable - it would require a larger refactor, but maybe we can make it part of the parent class? It would emphasize that it's the CSVVisitor's job to report errors regarding missing headers or invalid data that isn't otherwise easily generalized just from reading a CSV.

### DIFF
--- a/common/src/python/inputs/csv_reader.py
+++ b/common/src/python/inputs/csv_reader.py
@@ -2,7 +2,7 @@
 
 import abc
 from abc import ABC, abstractmethod
-from csv import DictReader, Error, Sniffer
+from csv import DictReader, Error
 from typing import Any, Dict, List, Optional, TextIO
 
 from outputs.errors import (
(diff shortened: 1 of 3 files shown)


bjkeller:
Thanks for pointing out the pattern for the error writer. I'd rather not add it to the abstract visitor class, but perhaps there is a different strategy

bjkeller:
This looks fine. I'm going to merge it.

PR #142, opened 2024-12-18
Identifier gears update

hewanadun:
Update identifier-lookup and identifier-provisioning gears
- Change enrollment session/acquisition labels and filename
- Change `identifier_mode` gear config to `database_mode`

### DIFF
--- a/common/src/python/enrollment/enrollment_subject.py
+++ b/common/src/python/enrollment/enrollment_subject.py
@@ -5,6 +5,7 @@ from typing import Optional
 
 from flywheel_adaptor.subject_adaptor import SubjectAdaptor
 from identifiers.model import GUID_PATTERN, NACCID_PATTERN
+from keys.keys import DefaultValues
 from pydantic import BaseModel, Field, ValidationError
 
 from enrollment.enrollment_transfer import (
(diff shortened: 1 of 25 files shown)


common/src/python/enrollment/enrollment_subject.py
```diff
@@ -96,17 +97,18 @@ def update_enrollment_info(self, record: EnrollmentRecord) -> None:
...
+        formdate = record.start_date.strftime("%Y-%m-%d")
+        session_label = f'{DefaultValues.ENRL_SESSION_LBL_PRFX}{formdate}'
         self.upload_acquisition_file(
-            session_label='enrollment_transfer',
-            acquisition_label='enrollment',
-            filename='enrollment.json',
+            session_label=session_label,
+            acquisition_label=DefaultValues.ENROLLMENT_MODULE,
+            filename=
+            f'{record.naccid}_{session_label}_{DefaultValues.ENROLLMENT_MODULE}.json',
```
bjkeller, line 111: if we are going to use this naming scheme as a convention, probably should make a function to create the name

gear/identifier_lookup/src/python/identifier_app/main.py
```diff
@@ -52,15 +59,19 @@ def __get_writer(self):
...
 
-        If the header doesn't have `ptid`, returns an error.
+        If the header doesn't have `ptid` or `adcid`, returns an error.
 
         Args:
           header: the list of header names
         Returns:
           True if `ptid` occurs in the header, False otherwise
         """
         if FieldNames.PTID not in header:
```
bjkeller, line 69: Not sure you need to change this now, but I think we should try to avoid enumerating conditionals. 

I had written the linked code to avoid repeated conditionals
https://github.com/naccdata/flywheel-gear-extensions/blob/0f4f8c6ae3bf03bbb5f31054af00fb1012594ad3/gear/identifier_provisioning/src/python/identifier_provisioning_app/main.py#L416

gear/identifier_lookup/src/python/identifier_app/main.py
```diff
@@ -82,7 +93,14 @@ def visit_row(self, row: Dict[str, Any], line_num: int) -> bool:
         Returns:
           True if there is a NACCID for the PTID, False otherwise
         """
-        writer = self.__get_writer()
+
+        if int(row.get(FieldNames.ADCID, -1)) != self.__adcid:
```
bjkeller, line 97: would it be possible to use enrollment.enrollment_transfer.CenterValidator here?

bjkeller (changes requested):
I think the only thing that needs to be done is remove the change wrapping the `info` object for transfer deserialization.

The other two notes are about not duplicating code, and trying to set a convention on checking headers

bjkeller (approved):
Looks good. Thanks for making those changes

PR #150, opened 2025-01-05
Error metadata updates

hewanadun:
Move error metadata to visit error log files stored at project level. 
Error logs are named as <ptid>_<visit date>_<module>_qc-status.log
All the gears store metadata in the visit error logs for centralized error reporting
Visit log file content will have the cumulated history of qc status over the time for a particular visit. However, the log file metadata will only be for the latest submission of that visit.

This turned out to be a bigger PR than I expected. It would be easy check gear by gear and in the order of execution in the pipeline.

For form submissions
identifier-lookup -> form-transformer -> form-qc-coordinator -> form-qc-checker

For enrollment submissions
form-qc-checker -> identifier-provisioning

### DIFF
--- a/common/src/python/enrollment/enrollment_transfer.py
+++ b/common/src/python/enrollment/enrollment_transfer.py
@@ -91,7 +91,7 @@ class EnrollmentRecord(BaseModel):
     transfer_to: Optional[TransferRecord] = None
 
     def query_object(self) -> IdentifierQueryObject:
-        """Creates an object for creating identifiers in the respository.
+        """Creates an object for creating identifiers in the repository.
 
         Returns:
           the identifier query object to use in batch identifier creation
@@ -107,7 +107,7 @@ def has_value(row: Dict[str, Any], variable: str, value: int) -> bool:
     Args:
       row: dictionary for a data row
       variable: the variable name
-      value: the expected valute
+      value: the expected value
     Returns:
       True if the variable has the value. False, otherwise.
     """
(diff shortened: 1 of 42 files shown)


echeng06 (dismissed):
Have some comments more related to the design/workflow, otherwise looks good to me! But would be good to also get Ben's eyes on it. 

- Not sure if we should account for race conditions, but I can see instances where it happens if we're not careful (especially from user error, like they upload multiple bulk visits with conflicting data), e.g. multiple gears trying to update the error log at the same time or writing errors that conflicts with another gear, etc.
- Could be refactored further to bundle some fields only passed specifically to write the error metadata (e.g. pass/fail, gear name, error log template, error log name, etc.) in a data object of sorts to reduce some of the complexity

hewanadun:
> * Not sure if we should account for race conditions, but I can see instances where it happens if we're not careful (especially from user error, like they upload multiple bulk visits with conflicting data), e.g. multiple gears trying to update the error log at the same time or writing errors that conflicts with another gear, etc.

Yes, we need to account for race conditions (still thinking about a way to handle multiple concurrent submissions), we can discuss.

bjkeller (dismissed):
Looks good. Please go ahead and merge

bjkeller:
@hewanadun @echeng06 I think maybe we should increment version numbers at least at minor level for these changes

hewanadun:
> @hewanadun @echeng06 I think maybe we should increment version numbers at least at minor level for these changes

Yes, updating docs and version numbers

hewanadun:
@echeng06 could you please take a look at the identifier-provisioning gear changes (sorry for not including this in the initial round).

echeng06 (approved):
Looks good to me!

PR #151, opened 2025-01-05
Add centerid lookup

bjkeller:
Refactors and extends the identifier lookup gear to allow looking up ADCID-PTID pair from NACCID.
This enables splitting a CSV file with just NACCID across centers.

### DIFF
--- a/gear/identifier_lookup/src/python/identifier_app/main.py
+++ b/gear/identifier_lookup/src/python/identifier_app/main.py
@@ -4,6 +4,11 @@ import logging
 from typing import Any, Dict, List, Optional, TextIO
 
 from enrollment.enrollment_transfer import CenterValidator
+from gear_execution.gear_execution import GearExecutionError
+from identifiers.identifiers_repository import (
+    IdentifierRepository,
+    IdentifierRepositoryError,
+)
 from identifiers.model import IdentifierObject
 from inputs.csv_reader import CSVVisitor, read_csv
 from keys.keys import FieldNames
@@ -17,7 +22,7 @@ from outputs.outputs import CSVWriter
 log = logging.getLogger(__name__)
 
 
-class IdentifierVisitor(CSVVisitor):
+class NACCIDLookupVisitor(CSVVisitor):
     """A CSV Visitor class for adding a NACCID to the rows of a CSV input.
 
     Requires the input CSV has a PTID column, and all rows represent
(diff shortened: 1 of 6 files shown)


bjkeller:
Writing tests for the NACCID -> ADCID-PTID mapping would require mocking the identifier registry.
Letting it slide at the moment

gear/identifier_lookup/src/python/identifier_app/main.py
```diff
@@ -111,8 +115,91 @@ def visit_row(self, row: Dict[str, Any], line_num: int) -> bool:
...
+        """Prepares the visitor to write a CSV file with the given header.
+
+        If the header doesn't have `naccid`, returns an error.
+
+        Args:
+          header: the list of header names
+        Returns:
+          True if `naccid` occurs in the header, False otherwise
+        """
+        expected_columns = {FieldNames.NACCID}
```
echeng06, line 156: Not sure if we also need to allow for different cases of NACCID here (e.g. `NACCID` vs `naccid`)

gear/identifier_lookup/src/python/identifier_app/main.py
```diff
@@ -111,8 +115,91 @@ def visit_row(self, row: Dict[str, Any], line_num: int) -> bool:
...
+        if not set(expected_columns).issubset(set(header)):
+            self.__error_writer.write(missing_field_error(expected_columns))
+            return False
+
+        self.__header = header
+        self.__header.append(FieldNames.ADCID)
```
echeng06, line 162: Think you also need to append PTID here. And should we check for ADCID/PTID already being in the header or can we assume they won't be there?

echeng06 (changes requested):

hewanadun:
@bjkeller These changes may cause merge conflicts with the changes for error reporting in identifier-lookup gear. Can you take a look at that PR and see which one we should merge first.

gear/identifier_lookup/src/python/identifier_app/main.py
echeng06: Also in the manifest, the description for `input_file` is still "The CSV input file. Each row must have ADCID and PTID"; can remove the second part.

bjkeller: I'm going to assume they aren't there

bjkeller: also changed the manifest

echeng06 (dismissed):
Looks good!

PR #154, opened 2025-01-09
Add Form Screening and Form Scheduler gears

echeng06:
Adds the Form Screening and Form Scheduler gears. Apologies for another huge PR, it was easier to develop these together. Things also may be a bit messy as I was just trying to get things to work first.

Main changes:
* Adds the two gears
* Refactors code out of `form_qc_coordinator` to `common/gear_execution/gear_trigger.py` and `common/jobs/job_poll.py`, which are related to triggering gears and polling jobs, respectively


hewanadun (changes requested):
- In `JobPoll.is_job_complete` why not use `poll_job_status_by_id` instead of `get_job_by_id` and `poll_job_status`?
- For GearConfigs it might be better to define sub classes for each gear (loading the file as it is allows more flexibility but we wouldn't catch if we update the manifest and forgot to update the configs file) - not urgent, can be added later.
- In `FormSchedulerQueue.add_files` and `send_email` methods project is defined as `ProjectOutput`, change this to `Project` for type compatibility
- In `FormSchedulerQueue.add_files` only add `.csv` files (skip `.json`)
- Noticed there are some typos in manifests, docs, and docstrings, please fix
- Implementation of form_scheduler gear `wait_for_submission_pipeline` is not correct, currently it only waits for the first match that returns by `find_job` to complete.
   - Add a `find_jobs` method to proxy class that returns the list of jobs matches the search string (there's a jobs.find() method in FW)
   - Sleep for `t` seconds and poll again, `t` can be adjusted at each iteration depending on the number of jobs returned in the search.
   - Iterate until no matches found
- In form_screening gear use `context.metadata.add_file_tags` method to tag the file (it handles duplicates, you can find examples in other gears)

echeng06:
@hewanadun Thanks for reviewing!

1. Good point - I think I had copied that particular function over from `form-qc-coordinator` but then added `JobPoll.poll_job_status_by_id` later for something else and neglected to refactor it. Fixed!
2. Sounds good, I think I'm gonna put this in a followup PR just to get this huge monster of a PR in first
3. I have to use ProjectOutput else it fails `pants check` (e.g. `project.files` is invalid), also I think the type really is ProjectOutput?
5. Done
6. Done although I didn't see many so I might just be blind to the ones you're referring to?
7. Discussed over Teams, added a `find_jobs` method anyways since it might be useful later
8. Done

My plan is to go ahead and merge this in, address 2) in a separate PR, then test all 3 affected gears on sandbox to make sure everything looks good and I didn't break anything. If it all looks good there then I'll also update prod.

hewanadun (approved):
Thanks for making the changes. Yes, we can address (2) later.
For (3), actual type is `ProjectOutput` but `Project` is an alias to that. I don't think we have directly used `ProjectOutput` or `FileOutput` anywhere else (Project and FileEntry are generally used and defined in mypy-stubs). Not sure why `pants check` fail on that, it's OK if you can't figure it out, we can merge as it is and figure it out later.

echeng06:
@hewanadun Ahh I see - I think `files` just isn't in the Project stub file. I'll try to resolve that in the next PR, thanks!

PR #161, opened 2025-01-14
Preprocessing checks

hewanadun:
Add pre-processing checks


common/src/python/datastore/forms_store.py
```diff
@@ -0,0 +1,215 @@
...
+        ]
+
+        if extra_columns:
+            for extra_lbl in extra_columns:
+                extra_col = f'{DefaultValues.FORM_METADATA_PATH}.{extra_lbl}'
+                columns.append(extra_col)
+
+        filters = f'acquisition.label={module}'
+        if not find_all:
+            filters += f',{search_col}{search_op}{search_val}'
```
echeng06, line 174: This needs this statement added

```
            if search_op == DefaultValues.FW_SEARCH_OR:
                search_val = f"[{','.join(search_val)}]"
```

For whatever reason, the following two filters are different, and while they'll both match "I" the first one seems to not match "I4" (I think something about the space, since `[I, I4]` also doesn't work)

```
acquisition.label=UDS,file.info.forms.json.packet=|['I', 'I4']
acquisition.label=UDS,file.info.forms.json.packet=|[I,I4]
```

common/src/python/datastore/forms_store.py
```diff
@@ -0,0 +1,215 @@
...
+            Optional[List[Dict[str, str]]]: _description_
+        """
+        return self.__query_project(project=self.__ingest_project,
+                                    subject_lbl=subject_lbl,
+                                    module=module,
+                                    search_col=search_col,
+                                    search_val=search_val,
+                                    search_op=search_op,
+                                    qc_gear=qc_gear,
+                                    extra_columns=extra_columns)
```
echeng06, line 72: `find_all` is not getting passed so here

common/src/python/datastore/forms_store.py
```diff
@@ -0,0 +1,215 @@
...
+            return None
+
+        return self.__query_project(project=self.__legacy_project,
+                                    subject_lbl=subject_lbl,
+                                    module=module,
+                                    search_col=search_col,
+                                    search_val=search_val,
+                                    search_op=search_op,
+                                    qc_gear=qc_gear,
+                                    extra_columns=extra_columns)
```
echeng06, line 110: `find_all` also not getting passed here

echeng06 (changes requested):
Hey Chandima, I know this PR is still in draft but just noting some issues I've found while using the same code for the legacy sanity check gear - I fixed them there but I think this one needs to get out before that one.

common/src/python/keys/keys.py
```diff
@@ -55,6 +55,10 @@ class DefaultValues:
     MAX_POOL_CONNECTIONS = 50
     PROV_SUFFIX = 'provisioning'
     IDENTIFIER_SUFFIX = 'identifiers'
+    FW_SEARCH_OR = '=|'
+    FORM_METADATA_PATH = 'file.info.forms.json'
+    UDS_I4_PACKET = 'I4'
+    UDS_I_PACKET = 'I'
```
echeng06, line 61: I don't think these really qualify as "default" values; we may want to separate some stuff out to module-specific keys (so `C2TMODE`, `LBD_SHORT_VER`, `UDS_I4_PACKET`, `UDS_I_PACKET`, etc.).

Similarly stuff like `FORM_METADATA_PATH` might also fall better under `MetadataKeys`, and `FW_SEARCH_OR` could be part of something else that also defines all the `SearchOperators` defined in `forms_store.py`

common/src/python/preprocess/preprocessor.py
```diff
@@ -0,0 +1,536 @@
...
+        """
+
+        packet = input_record[FieldNames.PACKET]
+        if (packet not in module_configs.initial_packets
+                and packet not in module_configs.followup_packets):
+            log.error('%s - %s/%s',
+                      preprocess_errors[SysErrorCodes.INVALID_PACKET], module,
+                      packet)
+            self.__error_writer.write(
+                preprocessing_error(
```
echeng06, line 79: I think we can refactor a lot of this to utilize Pydantic model(s) for query/error arguments in order to avoid passing what is nearly the same arguments over and over to `FormsStore.query_form_data` and `errors.preprocessing_error` (something like [this](https://github.com/naccdata/flywheel-gear-extensions/blob/feature/legacy-sanity-checks/gear/legacy_sanity_check/src/python/legacy_sanity_check_app/main.py#L62)) since it's quite dense to read otherwise but we can deal with that later.

echeng06 (approved):
Thanks for making those changes! Looks good to me for now, there are several spots I think we can be refactor but can be dealt with later.

hewanadun:
@echeng06 Yes, those are good points. I'm considering moving the pre-processing checks to a JSON definition, so will address those in a future PR. Going to merge this for now.

PR #165, opened 2025-01-15
Refactor getting parent from file

echeng06:
We don't have to merge this right away since a lot of the affected gears are in active development for UDSv4 (I also left form-transformer and form-qc-checker alone), but I noticed this specific logic was super redundant so wanted to get it refactored.

The actual `InputFileWrapper.get_parent_project` call was already merged to main since I need it for another new gear.


hewanadun (dismissed):
- Can use `ProjectAdaptor.create` to get the ingest_project
- In `LegacySanityChecker.run_all_checks` need to run it only for the module of the current file, can get the current module name from the acquisition label or file metadata (file.info.forms.json.module)
- Don't really need to run the checks on legacy project since these are uploaded after SAS checks passed (but no harm in leaving it as a sanity check to see whether there are multiple IVPs)
- Check on `num_legacy == 0`, this is not possible (If it gets to this stage that means SAS checks passed so an IVP packet must exist)
- There is a DefaultValue defined for UDS label
- When checking for I4, it is not necessary to retrieve the actual record using `get_visit_data` since PACKET value is already included in search result
- In `check_duplicate_visit` ingest results and retro results incorrectly labeled
- For gear template, we would want the default to be using `ContextClient`, `GearBotClient` is only for the gears that need credential access

echeng06:
@hewanadun I think you meant this for https://github.com/naccdata/flywheel-gear-extensions/pull/168 but updated!

hewanadun (approved):
Looks good!

PR #168, opened 2025-01-22
Legacy sanity checks gear

echeng06:
Adds gear to perform legacy sanity checks:

1. Checks that there are not multiple IVP packets across ingest and retrospective projects and modules, except for the case of a single I packet in UDSv3 and a single I4 packet in UDSv4 for the UDS module
2. Checks that there are no duplicate records across UDSv3 and UDSv4 projects (matching on packet, visit number, and visit date)

Changes to common code:

* Adds support for sending raw emails (e.g. for stuff that doesn't really need templating, like this) in `notifications.email`
* Adds `forms_store.FormQueryArgs` to help make passing query arguments to `FormsStore` a little easier
    * Lets `search_val` and `search_op` be optional if using `find_all`
* Adds `ProjectAdaptor.get_subject_by_id` to get a subject by its ID
* Updates some stuff in `templates/gear`for creating new gears

Also email ends up just looking like this, I don't think we need something fancy for these internal emails
image


hewanadun (approved):
Thanks for the updates!
I did some minor refactoring and also added the gear tag to input file.

PR #169, opened 2025-01-22
Refactor gear configs

echeng06:
For QC Coordinator and Form Screening; refactors GearConfigs to be more specific to the gear they are triggering.


hewanadun (dismissed):
In `GearInfo.load_from_file` it's kind of confusing to have `configs_file_path` attribute to be str or dict. It seems you added it for the ease of writing test cases. Can we load the test configs from an example file? [we can leave it as it is if it's too much work to set that up]

FYI - I fixed some ruff linter warnings.

PR #170, opened 2025-01-24
Module validation updates

hewanadun:
To validate a FTLD or LBD visit, it is required to have an approved UDS visit
- Modify the form-qc-coordinator to lookup matching UDS visit and pass it to form-qc-checker
- Modify form-qc-checker to load the supplement visit schema and merge the input record with supplement visit record


gear/form_qc_coordinator/src/python/form_qc_coordinator_app/coordinator.py
```diff
@@ -162,28 +168,89 @@ def update_last_failed_visit(self, file_id: str, filename: str,
...
+            f'{DefaultValues.FORM_METADATA_PATH}.{self.__module_configs.date_field}'
+        )
+        visitnum_key = (
+            f'{DefaultValues.FORM_METADATA_PATH}.{FieldNames.VISITNUM}')
 
         # sort the visits in the ascending order of visit date
         sorted_visits = sorted(visits, key=lambda d: d[date_col_key])
         visits_queue = deque(sorted_visits)
 
+        supplement_module = (self.__module_configs.supplement_module if
```
echeng06, line 251: This can just say `supplement_module = self.__module_configs.supplement_module`; the `if` part isn't really necessary since line 274 just checks it again.

gear/form_qc_checker/src/python/form_qc_app/definitions.py
```diff
@@ -40,13 +40,56 @@ def __init__(self,
...
+        """Append supplement schema to the given schema. Only assign the type
+        and set nullable to True, any other rules defined in the supplement
+        schema are skipped.
+
+        Args:
+            schema: schema for input visit data
+            supplement: schema for supplement module visit data
+        """
+        for field in supplement:
+            if field not in schema:
```
echeng06, line 79: Not sure if we want to throw an error or at least give a warning if the two schemas somehow have overlapping variables? They should be mutually exclusive so would imply an error on one of them, but not a big deal to leave as-is.

echeng06 (commented):
Overall logic makes sense and looks good to me, just left some nitpick comments!

Also not sure if we want to at some point reorganize where we're keeping all of our gear/form configs since we seem to be developing more and more. They currently live in the gear that most immediately uses it but it can be a bit confusing if 1) another gear wants to use it or 2) the gear that it specifies changes its manifest, you basically have to remember to update both the config and manifest together.

I imagine we could set up a Github action or similar that scrapes all gear manifests and converts them to the config file and dumps them into a singular location, maybe on pushes to main (form configs would still need to be manually written but can also live there). But this is a down-the-line and nice-to-have kind of thought, don't have to worry about it right now.

gear/form_qc_checker/src/python/form_qc_app/definitions.py
hewanadun: They are not mutually exclusive, header variables are common and we need to keep those rules for the current module.

echeng06: Ah right forgot about those, sounds good.

echeng06 (dismissed):

PR #179, opened 2025-02-06
Identifier Lookup: Fix case bug

echeng06:
Fixes issue where `CenterLookupVisitor.visit_row` was setting the keys to lowercase, causing `writer.write()` to fail with non-lower case headers

If we want to enforce lowercase keys, I can update this to allow for a `preserve_case` configuration value, which when set to False (the default), sets everything to lowercase. 

EDIT: Updated to pass `preserve_case` to `read_csv`, which will convert all keys to lowercase and replace spaces with underscores (and remove trailing whitespace). Default is set to `true` so backwards compatibility is conserved, and for something that changes data that should probably be an intentionally-passed parameter.

### DIFF
--- a/common/src/python/inputs/csv_reader.py
+++ b/common/src/python/inputs/csv_reader.py
@@ -47,7 +47,8 @@ def read_csv(*,
              visitor: CSVVisitor,
              delimiter: str = ',',
              limit: Optional[int] = None,
-             clear_errors: Optional[bool] = False) -> bool:
+             clear_errors: Optional[bool] = False,
+             preserve_case: bool = True) -> bool:
     """Reads CSV file and applies the visitor to each row.
 
     Args:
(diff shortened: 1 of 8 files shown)


hewanadun:
@echeng06  All JSON schemas are defined in lowercase, so CSV headers are expected to be in lowercase (this may not be applicable to this particular gear). I think it is better to convert the header to lowercase before any validation in the `csv_reader.py` `read_csv()` method?

echeng06:
@hewanadun This is more for loading in external CSVs (like the availability file from NIAGADS) - I'm not opposed to forcing lowercase there but having it as a config option would give us some flexibility for other use cases.

hewanadun (approved):
~~@echeng06 in `read_csv()` can we set `preserve_case: bool = False` as the default value, since for most of the gears, headers expected to be in lowercase?~~
On second thought, let's keep it as it is (since spaces are replaced with _ too, better to be intentional on that).

Also, please rebuild the gears and test the pipelines in sandbox before merge.

PR #183, opened 2025-02-07
CSV Subject Splitter tweaks

echeng06:
* Updates `hierarchy_labels` config to be type `string` instead of type `object`
* Updates `uploads.uploader.JSONUploader` to accept a `allow_updates` parameter (defaults `false`) and, if `false`, throw an error if the subject already exists
* Removes `assert parameter_store` statement


bjkeller (approved):
Looks good. Thanks for catching the json string config argument

PR #187, opened 2025-02-21
CSV Subject Splitter: Query subject first

echeng06:
Change query subject order by first searching, then adding if it does not exist. In the long run this should result in less API calls (also kind of backwards to act on a caught exception to begin with).


bjkeller (changes requested):
See the comments, but please
1. remove the allow_updates flag
2. change the add_subject method so that it calls find_subject before creating a new one

bjkeller (approved):
Looks good. Thank you!

PR #188, opened 2025-02-21
CSV Center Splitter: include exclude params

echeng06:
* Adds `include` and `exclude` config options to include/exclude centers as needed

### DIFF
--- a/gear/csv_center_splitter/src/docker/manifest.json
+++ b/gear/csv_center_splitter/src/docker/manifest.json
@@ -2,7 +2,7 @@
     "name": "csv-center-splitter",
     "label": "CSV Center Splitter",
     "description": "Splits a CSV by center ID",
-    "version": "0.1.1",
+    "version": "0.1.2",
     "author": "NACC",
     "maintainer": "NACC <nacchelp@uw.edu>",
     "cite": "",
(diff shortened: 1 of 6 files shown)


gear/csv_center_splitter/src/docker/manifest.json
```diff
@@ -56,6 +56,16 @@
             "type": "string",
             "default": ""
         },
+        "include": {
+            "description": "Comma-deliminated list of ADCIDs to include in the split. Will ignore all others.",
```
bjkeller, line 60: spelling: delimited

gear/csv_center_splitter/src/python/csv_center_splitter_app/run.py
```diff
@@ -29,14 +30,22 @@ def __init__(self,
                  adcid_key: str,
                  target_project: str,
                  staging_project_id: Optional[str] = None,
+                 include: Optional[List[str]] = None,
+                 exclude: Optional[List[str]] = None,
                  delimiter: str = ',',
                  local_run: bool = False):
         super().__init__(client=client)
 
+        if (include and exclude and set(include) & set(exclude)):
```
bjkeller, line 39: Please use the intersection method instead of the `&` operator.

The python documentation suggests that use of the operator can be error prone. Not necessarily as you have used it, but if someone changes something in the future

bjkeller (changes requested):
Looks good, but please change use of `&` to explicit call to `intersection` method

bjkeller (approved):

PR #189, opened 2025-02-23
Form Transformer: Add tests

echeng06:
This was getting hard to test manually so I tried directly unit testing it. Required some mocking of Flywheel objects/queries which is pretty crude/messy at the moment, but maybe can be refined if we add this kind of thing for other gears

Also fixes a few immediate things from the [original PR](https://github.com/naccdata/flywheel-gear-extensions/pull/186):

* Moved `transformed_records` to inside `CSVTransformVisitor`
* Under `common.errors.update_error_log_and_qc_metadata` there is an issue where if validation error data already exists in that same ptid/visitdate, it will overwrite it with the current record's errors (since the files are mapped by ptid/visitdate). Updated it so that we always append the errors instead
    * You can simulate this by submitting two records with incorrect modules but the same ptid/visitdate -  before the fix it would only report the latest record as having an issue
* `CSVTransformVisitor.__report_duplicates_within_current_batch` issue, see [this](https://github.com/naccdata/flywheel-gear-extensions/pull/186#discussion_r1966679666) comment
* In `CSVTransformVisitor.process_current_batch`, DIFF_VISITDATE and LOWER_VISITNUM errors being written to wrong (?) file, see [this](https://github.com/naccdata/flywheel-gear-extensions/pull/186#discussion_r1966686376) comment

### DIFF
--- a/common/src/python/outputs/errors.py
+++ b/common/src/python/outputs/errors.py
@@ -452,7 +452,8 @@ def update_error_log_and_qc_metadata(*,
                                      gear_name: str,
                                      state: str,
                                      errors: List[Dict[str, Any]],
-                                     reset_metadata: bool = False) -> bool:
+                                     reset_metadata: bool = False,
+                                     append_errors: bool = False) -> bool:
     """Update project level error log file and store error metadata in
     file.info.qc.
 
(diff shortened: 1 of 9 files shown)


hewanadun (approved):
Looks good! Thanks for writing these tests.

PR #192, opened 2025-02-24
Refactor/change subject splitter execution

bjkeller:
Changes CSV splitting for subjects so that it uploads a file for each row rather than accumulating rows

### DIFF
--- a/gear/csv_subject_splitter/src/python/csv_app/main.py
+++ b/gear/csv_subject_splitter/src/python/csv_app/main.py
@@ -1,8 +1,7 @@
 """Defines CSV to JSON transformations."""
 
 import logging
-from collections import defaultdict
-from typing import Any, DefaultDict, Dict, List, TextIO
+from typing import Any, Dict, List, TextIO
 
 from flywheel_adaptor.flywheel_proxy import ProjectAdaptor
 from inputs.csv_reader import CSVVisitor, read_csv
@@ -12,7 +11,7 @@ from outputs.errors import (
     empty_field_error,
     missing_field_error,
 )
-from uploads.uploader import JSONUploader, UploadTemplateInfo
+from uploads.uploader import JSONUploader, UploaderError, UploadTemplateInfo
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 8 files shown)


bjkeller:
Reopening. FW had indicated that there might be a REST endpoint that might be useful here. I canceled just incase it would change this logic. However, it would just affect the uploader, and so these changes still apply.

gear/csv_subject_splitter/src/python/csv_app/main.py
```diff
@@ -68,8 +67,13 @@ def visit_row(self, row: Dict[str, Any], line_num: int) -> bool:
                 empty_fields, line_num))
             return False
 
-        subject_lbl = row[FieldNames.NACCID]
-        self.__records[subject_lbl].append(row)
+        subject = self.__project.add_subject(row[FieldNames.NACCID])
```
echeng06, line 70: Another way we could save on API calls here would be to cache which subjects we've already added/looked up (just keep a mapping of NACCID to SubjectAdaptor). Would need to test to see how much it helps though.

gear/csv_subject_splitter/src/python/csv_app/main.py
```diff
@@ -94,24 +98,14 @@ def run(*, input_file: TextIO, destination: ProjectAdaptor,
...
-                            template_map=template_map,
-                            environment=environment)
-    upload_status = uploader.upload(subject_record_map)
-    if not upload_status:
-        notify_upload_errors()
-
-    return result and upload_status
+                      visitor=CSVSplitVisitor(req_fields=[FieldNames.NACCID],
+                                              project=destination,
+                                              uploader=JSONUploader(
```
echeng06, line 105: Not a huge fan of nesting object instantiation since it's hard to read, but functionally it's okay.

gear/csv_subject_splitter/test/python/test_csv_split_visitor.py
```diff
@@ -83,18 +86,49 @@ def nonvisit_data_stream(valid_nonvisit_table):
     yield stream
 
 
+class MockUploader(JSONUploader):
```
echeng06, line 89: I actually did something really similar here for another gear (https://github.com/naccdata/flywheel-gear-extensions/pull/189), and started putting it under `common.test_mocks`. Doesn't have to be in this PR/where we actually put it is flexible but at some point should definitely consolidate!

echeng06 (approved):
Overall looks good! Main suggestion is locally caching which subjects have been added/found already.

PR #197, opened 2025-02-28
APOE Transformer: Add a1/a2 encoding to output

echeng06:
Adds A1/A2 encoding to output.

On that note I realized it is keeping all input headers, so if NCRAD ever gives additional extra fields those will be propagated to the output.


bjkeller (approved):
Looks good.

Basically this gear isn't so much transforming as inserting the `apoe` column, so I think it is fine if any extra fields remain

PR #198, opened 2025-03-03
CSV Center Splitter: Add scheduling

echeng06:
Updates CSV Center Splitter to batch/schedule the centers. Will wait for the current batch to complete all of their project's downstream gears before uploading the next batch.

### DIFF
--- a/gear/csv_center_splitter/src/python/csv_center_splitter_app/main.py
+++ b/gear/csv_center_splitter/src/python/csv_center_splitter_app/main.py
@@ -168,9 +168,8 @@ def run(*,
         delimiter: The CSV's delimiter; defaults to ','
     """
     # split CSV by ADCID key
-    centers: Set[str] = include if include else set()
     visitor = CSVVisitorCenterSplitter(adcid_key=adcid_key,
-                                       include=centers,
+                                       include=include if include else set(),
                                        error_writer=error_writer)
     success = read_csv(input_file=input_file,
                        error_writer=error_writer,
@@ -185,6 +184,8 @@ def run(*,
             log.error(x['message'])
         return
 
+    # filter to centers that were actually found
+    centers = [adcid for adcid in visitor.split_data]
     project_map = generate_project_map(proxy=proxy,
                                        centers=centers,
                                        target_project=target_project,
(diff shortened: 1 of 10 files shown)


gear/csv_center_splitter/src/python/csv_center_splitter_app/main.py
```diff
@@ -200,29 +204,49 @@ def run(*,
             f"Missing {target_project} projects for the following " +
             f"ADCIDs: {missing_projects}")
 
-    log.info(f"Writing split results for the following ADCIDs: {centers}")
+    if not batch_size:  # just make one big chunk
+        batch_size = len(visitor.split_data)
```
bjkeller, line 208: At this point, it should actually be an error if there is no batch size. So make the param non-optional

gear/csv_center_splitter/src/python/csv_center_splitter_app/main.py
```diff
@@ -200,29 +204,49 @@ def run(*,
...
+                f"ADCID {adcid} with project ID {project.id}")  # type: ignore
+
+            contents = write_csv_to_stream(headers=visitor.headers,
+                                           data=data).getvalue()
+            file_spec = FileSpec(name=filename,
+                                 contents=contents,
+                                 content_type="text/csv",
+                                 size=len(contents))
+
+            if proxy.dry_run:
```
bjkeller, line 240: Please clean up the dry_run checks. Change this to something like 

```python
if proxy.dry_run:
    log.info(...)
    continue
```

and then take out the else and the other check

gear/csv_center_splitter/src/python/csv_center_splitter_app/main.py
```diff
@@ -145,6 +146,8 @@ def run(*,
         target_project: str,
         staging_project_id: Optional[str] = None,
         include: Optional[Set[str]] = None,
+        batch_size: Optional[int] = None,
```
bjkeller, line 149: remove Optional. we are requiring it at this point

gear/csv_center_splitter/src/python/csv_center_splitter_app/run.py
```diff
@@ -33,6 +33,8 @@ def __init__(self,
                  include: Set[str],
                  exclude: Set[str],
                  staging_project_id: Optional[str] = None,
+                 batch_size: Optional[int] = None,
```
bjkeller, line 36: let's make batch size non-optional. otherwise, known as required

bjkeller (changes requested):
Let's make the batch size a required, rather than Optional, parameter throughout.

Also, please simplify the dry_run logic in main.py

gear/csv_center_splitter/src/python/csv_center_splitter_app/main.py
echeng06: The other check is in the outer loop - I could have it act on `project_ids_list` instead just so we ensure we're not acting on no projects either (if somehow the batch logic screwed up).

bjkeller: Great. That works. thanks!

PR #209, opened 2025-03-31
Feature/add attribute curator

bjkeller:
Adds the attribute-curator gear using the nacc-attribute-deriver.
Uses multiprocessing over subjects


docs/uds-curator/index.md
```diff
@@ -0,0 +1,6 @@
+# UDS Curator
```
echeng06, line 1: Should update these docs to attribute-curator instead of uds-curator.

gear/attribute_curator/src/docker/BUILD
```diff
@@ -0,0 +1,9 @@
+file(name="manifest", source="manifest.json")
+
+docker_image(name="attribute-curator",
+             source="Dockerfile",
+             dependencies=[
+                 ":manifest",
+                 "gear/attribute_curator/src/python/attribute_curator_app:bin"
+             ],
+             image_tags=["0.0.1a", "latest"])
```
echeng06, line 9: Can do after approval/testing but should be updated to probably 0.0.1 or 0.1.0.

gear/attribute_curator/src/docker/manifest.json
```diff
@@ -0,0 +1,53 @@
...
+            "suite": "Curation",
+            "show-job": true
+        }
+    },
+    "inputs": {
+        "api-key": {
+            "base": "api-key",
+            "read-only": false
+        },
+        "derive_rules": {
```
echeng06, line 30: Don't think we need this after the changes made to attribute-deriver?

gear/attribute_curator/src/python/attribute_curator_app/run.py
```diff
@@ -0,0 +1,112 @@
...
+        Raises:
+          GearExecutionError if any expected inputs are missing
+        """
+
+        client = ContextClient.create(context=context)
+        proxy = client.get_proxy()
+
+        filename_pattern = context.config.get('filename_pattern', "*.json")
+
+        derive_rules = InputFileWrapper.create(input_name='derive_rules',
```
echeng06, line 58: Also need to remove logic here

echeng06 (changes requested):
Some comments to clean up deprecated code/docs but looks good to me otherwise, thanks!

common/src/python/curator/scheduling.py
```diff
@@ -0,0 +1,299 @@
...
+                ColumnModel(data_key='file.info.forms.json.visitdate',
+                            label="visit_date"),
+                ColumnModel(data_key="file.modified", label="modified_date"),
+                ColumnModel(data_key="file.info.raw.study_date",
+                            label="study_date"),
+                ColumnModel(data_key="file.info.raw.scan_date",
+                            label="scan_date"),
+                ColumnModel(data_key="file.info.raw.scandate",
+                            label="scandate"),
+                ColumnModel(data_key="file.info.raw.scandt", label="scandt")
```
echeng06, line 227: So I could be misremembering/not fully understanding how data views work, but I think there was an issue with the builder not returning anything if a column was empty. I'm not sure if this has been tested so just flagging in case it's a problem.

gear/attribute_curator/src/docker/manifest.json
bjkeller: removed

gear/attribute_curator/src/python/attribute_curator_app/run.py
bjkeller: removed

PR #212, opened 2025-04-08
Fix/center info key

bjkeller:
Changes key for the center info map to a string rather than an integer.
Also, adds methods to center info map class and updates code that was previously using the map directly.


bjkeller:
@echeng06 @hewanadun This touches several gears. I think I caught them all, but let me know if I missed any

hewanadun (approved):
Looks good!

echeng06 (approved):

PR #213, opened 2025-04-09
Attribute Curator: Fix silent failures

echeng06:
Fixes issue where curator fails silently on pooled exceptions and updates the deriver. 

* Updates `nacc_attribute_deriver` to `1.2.2`
* Removes rows from ViewResponseModel where all fields are null
* Fixes issue where multiprocessing was failing silently on exception


bjkeller (approved):
Looks good

PR #219, opened 2025-04-14
CenterManagement: Refactor tags

echeng06:
Remove tags from CenterInfo and refactor how tags are managed by center_management gear. I don't think tags are being used in this context anywhere else. 

From this old conversation here: https://github.com/naccdata/flywheel-gear-extensions/pull/128#discussion_r1878740614

Tagging as unreleased for now as it is not critical and can be bundled with the next necessary update for the Center Management gear.


bjkeller (approved):
Looks good

PR #225, opened 2025-04-24
Regression curation gear

echeng06:
Two main parts:

* Adds the Regression Curation gear, which refactors curation code
* Update Form Curator (Attribute Curator gear) to run a second pass over the files to apply cross-sectional values


common/src/python/curator/regression_curator.py
```diff
@@ -0,0 +1,141 @@
...
+            return
+
+        record = None
+        expected = subject.label
+        # derived_vars = None
+        if scope == 'uds':
+            #derived_vars = table.get('file.info.derived', None)
+            visitdate = table['file.info.forms.json.visitdate']
+            expected = f"{expected} {visitdate}"
+            for r in self.__baseline[subject.label]:
```
bjkeller, line 121: this reference has a type error because it can be None

Could you preprocess the data to index by visit date so that you don't have to do the linear searches?

common/src/python/curator/regression_curator.py
```diff
@@ -0,0 +1,141 @@
...
+                if visitdate == r['visitdate']:
+                    record = r
+                    break
+        else:
+            #derived_vars = table.get('subject.info.derived', None)
+            record = self.__baseline[subject.label][-1]
```
bjkeller, line 127: also has a type error because it can be None

common/src/python/curator/scheduling.py
```diff
@@ -334,9 +318,7 @@ def __compute_cores(self) -> int:
     def apply(self,
               context: GearToolkitContext,
               curator_type: Type[C],
-              deriver: AttributeDeriver,
-              curation_tag: str,
-              force_curate: bool = False) -> None:
+              curator_type_args: Optional[Dict[str, Any]] = None) -> None:
```
bjkeller, line 321: I'd rather have explicit arguments so that typing can help identify errors when there are some.

bjkeller (changes requested):
a couple of points:
1. There is a linear search for a visit date in the regression curator. Can you do some initial indexing to avoid that?
2. I'd like to avoid **args so that we can take advantage of type checking to catch bugs. In the cases you change, you replaced three explicit arguments with an args object. (and, long argument lists are probably a symptom of a different design problem)

common/src/python/curator/regression_curator.py
echeng06: Sure - I think I started this route to generalize regression testing against non-form data, but since it's effectively only looking at UDS/QAF we can assume there is a visitdate to index.

common/src/python/outputs/errors.py
```diff
@@ -434,6 +441,13 @@ def clear(self):
         self.__errors.clear()
 
 
+class MPListErrorWriter(ListErrorWriter):
+    """ListErrorWriter for multiprocessing."""
+
+    def __init__(self, container_id: str, fw_path: str) -> None:
+        super().__init__(container_id, fw_path, Manager().list())
```
bjkeller, line 448: I'll admit that I haven't looked to see where this is used, but is there a need for another class? Can you just make a ListErrorWriter and use the Manager there? (My thought is that this package doesn't need to be aware of the Manager type.)

bjkeller (changes requested):
just one minor detail in errors package: whether need MP specific list error

common/src/python/outputs/errors.py
echeng06: Updated! You are correct there isn't really a need for another class.

bjkeller (approved):
Thanks for fixing that picky little detail!

PR #230, opened 2025-05-13
Form screening updates

hewanadun:


gear/form_screening/src/python/form_screening_app/format.py
```diff
@@ -0,0 +1,67 @@
...
+        """Convert the header to lowercase. Remove any REDCap specific columns
+        form the header.
+
+        Args:
+          header: the list of header names
+
+        Returns:
+          True
+        """
+        self.__header = [column.strip().lower() for column in header]
```
echeng06, line 41: `read_csv` has a `preserve_case` argument that explicitly needs to be set to `False` but will handle lowercase of the headers + rows. It also converts to snakecase. We probably want to use that instead unless we need to avoid the snakecase part. 

In general this could also be simplified to
`self.__header = [column for column in header if column not in RedCapKeys]`

gear/form_screening/src/python/form_screening_app/format.py
```diff
@@ -0,0 +1,67 @@
...
+        """Remove any REDCap specific columns form the row.
+
+        Args:
+          header: the list of header names
+
+        Returns:
+          True
+        """
+
+        out_row = {
```
echeng06, line 59: Same as the comment in `visit_header`, in this case 

`out_row = {key: value, for key, value in row.items() if key not in RedCapKeys}`

gear/form_screening/src/python/form_screening_app/main.py
```diff
@@ -24,67 +30,130 @@ class FormSchedulerGearConfigs(GearConfigs):
 
 def run(*, proxy: FlywheelProxy, context: GearToolkitContext,
         file_input: InputFileWrapper, accepted_modules: List[str],
-        queue_tags: List[str], scheduler_gear: GearInfo,
-        error_writer: ListErrorWriter) -> bool:
+        queue_tags: List[str], scheduler_gear: GearInfo):
```
echeng06, line 33: Kind of a nitpick but this method is probably close to violating the complexity lint error. Would be good to split out into more functions, like maybe the part that writes the new file at least, or move more of the metadata context stuff back to `run.py`.

gear/form_screening/src/python/form_screening_app/main.py
```diff
@@ -24,67 +30,130 @@ class FormSchedulerGearConfigs(GearConfigs):
...
+                }
+            }
+        }
+    }
+
+    contents = out_stream.getvalue()
+    if len(contents) > 0:
+        log.info("Saving file %s in UTF-8", file_input.filename)
+        file_spec = FileSpec(name=file_input.filename,
+                             contents=str.encode(contents).decode(),
```
echeng06, line 117: Is there a need to explicitly encode/decode this?

gear/form_screening/src/python/form_screening_app/run.py
```diff
@@ -55,6 +55,22 @@ def create(
...
+        gear_name = context.manifest.get('name', 'form-screening')
+
+        # We save the formatted file with same name as input file
+        # To prevent gear rules running into a loop check whether the file is screened
+        time.sleep(15)
+        file_entry = file_input.file_entry(context=context)
+        if gear_name in file_entry.tags:
+            log.info("Input file %s already screened and formatted",
+                     file_entry.name)
+            exit(0)
```
echeng06, line 72: Probably want to use `sys.exit(0)`, there are some nuances with just using `exit(0)`.

The extra gear execution isn't ideal, but not sure how we can get around it with how gear rules work unless we rename the fixed file (which I'm guessing we don't want to do here). The only workaround I can think of is another pre-pre gear that renames or tags the file for form-screening, which also isn't really ideal (defeats the whole purpose of form-screening being the preprocessing gear). 

However I'm worried about the wait condition here - 15 seconds might not be enough. If form-screening gets stuck in pending then we could accidentally submit a whole bunch of gear instances. Is it possible to maybe instead check for if either a) the tag is there or b) there is already an existing form-screening job (pending/running) on this file in the same project? The b) case should help in the case of stuck gear jobs.

echeng06 (changes requested):
Mostly nitpicks but in particular see the one about using the 15 second sleep on re-execution.

hewanadun:
@echeng06 Thanks for the comments
- I agree that the extra gear execution isn't ideal but saving the file with different name or using a new gear makes things complex for ADRC portal UI and error reporting.
- Changed back to uploading the file using gear context, so we can check the origin of the file which is more reliable. I was able to get the tag/metadata update working with help from FW.
- I saw that `read_csv` has a `preserve_case` argument but we don't want the `snakecase` conversion

gear/form_screening/src/python/form_screening_app/run.py
```diff
@@ -55,6 +56,27 @@ def create(
...
+        gear_name = context.manifest.get('name', 'form-screening')
+
+        # We save the formatted file with same name as input file
+        # To prevent gear rules running into a loop check whether the file is screened
+        # Check file.origin.type to identify whether the file is generated by a gear job
+        file_entry = file_input.file_entry(context=context)
+        if file_entry.origin.type == 'job':
+            log.info("Skip processing input file %s generated by gear job %s",
+                     file_entry.name, file_entry.origin.id)
+            time.sleep(15)
```
echeng06, line 73: Is this sleep/check on tags still needed if we're just going to exit regardless?

echeng06 (dismissed):
One final nitpick but looks good otherwise! Using the file origin is definitely more reliable, thanks for making those changes.

gear/form_screening/src/python/form_screening_app/run.py
hewanadun: Left it to get an idea on how quickly the tag is added. Also, it might be helpful in troubleshooting.

echeng06: Okay sounds good. Approved!

PR #234, opened 2025-05-26
Add standalone forms

hewanadun:


echeng06 (approved):
Functionally this looks fine, however at some point we probably want to refactor the preprocessor to further modularize the code, especially to avoid passing so many parameters and if/else around since its resulting in a lot of bloat.

For example in the `preprocess` method, a lot of the same params are being passed to each check. We could maybe make a `RecordProcessor` object that takes in all of those as instance variables and call `record.check_whatever()` on it. It could further be split into processors for IVP, I4, and FVP forms since we know there is specific behavior/checks associated with each.

hewanadun:
@echeng06 Thanks for the feedback! Yes, I agree that preprocessor needs refactoring. Planning to do it in a separate PR.

PR #238, opened 2025-06-03
Update to utf-8-sig

echeng06:
Updates anywhere a file is read in from an external source to use `utf-8-sig` to handle BOM encoding.

### DIFF
--- a/gear/identifier_lookup/src/python/identifier_app/run.py
+++ b/gear/identifier_lookup/src/python/identifier_app/run.py
@@ -228,7 +228,8 @@ class IdentifierLookupVisitor(GearExecutionEnvironment):
             contents = out_file.getvalue()
             if len(contents) > 0:
                 log.info("Writing contents")
-                with context.open_output(filename, mode='w',
+                with context.open_output(filename,
+                                         mode='w',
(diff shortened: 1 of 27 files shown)


gear/identifier_lookup/src/python/identifier_app/run.py
```diff
@@ -228,8 +228,9 @@ def run(self, context: GearToolkitContext):
             contents = out_file.getvalue()
             if len(contents) > 0:
                 log.info("Writing contents")
-                with context.open_output(filename, mode='w',
-                                         encoding='utf-8') as fh:
+                with context.open_output(filename,
+                                         mode='w',
+                                         encoding='utf-8-sig') as fh:
```
hewanadun, line 233: I think we should keep the output file encoding as 'utf-8'?

hewanadun (approved):
Check the comment on output file in identifier_lookup gear. Rest looks good!

gear/identifier_lookup/src/python/identifier_app/run.py
echeng06: Yes, sorry I didn't mean to update writes. Fixed!

PR #246, opened 2025-06-25
Attribute Curator: 1.4.1 update

echeng06:
Main update is updating `nacc-attribute-deriver` [to the 1.4.1](https://github.com/naccdata/nacc-attribute-deriver/releases/tag/v1.4.1) release, but also includes some changes to optimize the runtime.

Attribute curator:
* Updates `nacc-attribute-deriver` to `1.4.1`
* Updates to handle MEDS (scope not yet used in 1.4.1 and ignored)
* Adds `debug` argument to reduce verbosity of logging statements
* Splits out scheduler models - updates curation `post_process` to take in `FileModel` instead of string `file_id` so it can access the filename without making an API call

Regression curator:
* Adds `debug` argument to reduce verbosity of logging statements
* Updates MQT regression testing to not be required

Lockfile update:
```
==                    Upgraded dependencies                     ==

  boto3                          1.38.43      -->   1.38.44
  boto3-stubs                    1.38.43      -->   1.38.44
  botocore                       1.38.43      -->   1.38.44
  nacc-attribute-deriver         1.4.0        -->   1.4.1
```

### DIFF
--- a/common/src/python/curator/curator.py
+++ b/common/src/python/curator/curator.py
@@ -94,7 +94,7 @@ class Curator(ABC):
 
         if (self.curation_tag and not self.force_curate
                 and self.curation_tag in file_entry.tags):
-            log.info(f"{file_entry.name} already curated, skipping")
+            log.debug(f"{file_entry.name} already curated, skipping")
             return
 
         scope = determine_scope(file_entry.name)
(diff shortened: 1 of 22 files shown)


bjkeller (approved):
This seems fine

PR #264, opened 2025-07-25
Eliminate catch exception

hewanadun:
Remove catching Exception and replace with handlers for more specific error types


common/src/python/curator/form_curator.py
```diff
@@ -80,7 +83,7 @@ def execute(
         """
         try:
             self.__deriver.curate(table, scope)
-        except Exception as e:
+        except AttributeDeriverError as e:
```
bjkeller, line 86: @echeng06 Does this seem OK? Can we make sure all exceptions from attribute deriver are re-raised as AttributeDeriverError?

echeng06: Sorry for getting to this so late - this is probably okay. The reason I had it generally catching any exceptions is because I didn't want a single edge case to completely halt execution. We just need to be careful about catching all exception in the attribute deriver.

common/src/python/inputs/csv_reader.py
```diff
@@ -100,8 +108,8 @@ def read_csv(
             success = row_success and success
             if limit and count >= limit:
                 break
-    except Exception as error:
-        error_writer.write(malformed_file_error(str(error)))
+    except CSVVisitorError as error:
+        log.error(f"Error occurred while reading CSV file: {error}")
```
echeng06, line 112: So I'm not sure if there is an advantage with explicitly throwing/catching CSVVisitorError vs just having `visit_row` return False. I think most of our current visitor classes just return False on malformed input. It might be better to just not have a try-catch here at all? CSVVisitorError might be better suited for an error that really does need to halt execution.

echeng06 (changes requested):

common/src/python/inputs/csv_reader.py
```diff
@@ -100,8 +108,8 @@ def read_csv(
             success = row_success and success
             if limit and count >= limit:
                 break
-    except Exception as error:
-        error_writer.write(malformed_file_error(str(error)))
+    except CSVVisitorError as error:
```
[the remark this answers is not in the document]
bjkeller: OK, I added a `visitor.valid_row()` predicate that is checked before the `visitor.visit_row()` call.
It set success to False and breaks the loop, which allows the error_writer calls to be made.

echeng06 (approved):
Looks good, thanks! Should be a different PR (I can put in a ticket) but probably want to comb through and detangle any visitor logic that should be moved to validate_row

PR #265, opened 2025-07-25
Support co-enrollment in affiliated studies

bjkeller:
* Adds ability to designate the enrollment strategy a center uses in an affiliated study: whether all participants are co-enrolled, or whether (some) participants are enrolled separately.
* Changes pipeline project creation so that they are not created for an affiliated studies in a center where the participants are all co-enrolled
* Also, includes refactoring of permission management.


bjkeller:
ok, I didn't do this correctly

PR #266, opened 2025-07-25
Feature/refine study model

bjkeller:
* Changes how studies are specified, making it possible to indicate for each center whether participant enrollment for an affiliated study is by co-enrollment or separate enrollment. Also adds flag for studies that have legacy data that will require ETL pipelines.
* Changes study pipeline creation for a study so that it avoids creating pipelines when participants are co-enrolled.
* Refactors permission management

### DIFF
--- a/common/src/python/projects/study_mapping.py
+++ b/common/src/python/projects/study_mapping.py
@@ -226,7 +226,7 @@ class AggregationMapper(StudyMapper):
         Args:
           center: the center group
         """
-        if not self.study.is_primary():
+        if not self.study.has_legacy():
             log.warning(
                 "Will not create retrospective projects for study %", self.study.name
             )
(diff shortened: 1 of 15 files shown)


common/src/python/projects/study_mapping.py
```diff
@@ -174,11 +226,23 @@ def __add_retrospective(self, center: CenterGroup) -> None:
         Args:
           center: the center group
         """
-        labels = [
-            f"retrospective-{datatype.lower()}" for datatype in self.__study.datatypes
-        ]
-        for label in labels:
-            center.add_project(label)
+        if not self.study.is_primary():
```
hewanadun, line 229: Don't we need retrospective projects if the affiliated study has separately enrolled participants with UDSv3 data?

hewanadun (commented):
Changes looks good. Please check the question on creating retrospective projects.

common/src/python/projects/study_mapping.py
bjkeller: Thanks for pointing this out. I am going to add an explicit "legacy" flag to indicate this instead of assuming only the primary study has legacy data.

PR #267, opened 2025-08-04
Refactor/error objects

bjkeller:


echeng06 (approved):
Looks good to me, thanks for doing this refactor!

This should be in a different refactor, but the main thing that jumps out as me is the code surrounding the error log file can probably be further refactored into a class. We currently pass the error log name around everywhere to do stuff so it's effectively an instance with extra steps at that point.

PR #276, opened 2025-08-08
Update error metadata

hewanadun:
Add date and naccid to error metadata
Fix error reporting for copying downstream gears metadata


bjkeller (approved):
Looks good. Thanks for adding those attributes

PR #277, opened 2025-08-11
Fix/group permissions for center group creation

bjkeller:
* Includes `audit` user role in group permissions, which fixes a downstream error that occurs when adding users to a center group.
* Restores tags to center info model and fixes related deserialization issue.


echeng06 (approved):
Looks good to me!

PR #278, opened 2025-08-14
Feature/qc initial visit

echeng06:


gear/form_qc_checker/src/python/form_qc_app/datastore.py
```diff
@@ -269,19 +336,47 @@ def get_previous_nonempty_record(
...
+        visit_data = self.__forms_store.get_visit_data(
+            file_name=self.__initial_visit["file.name"],
+            acq_id=self.__initial_visit["file.parents.acquisition"],
+        )
+
+        if self.__check_nonempty(ignore_empty_fields, visit_data):
+            return visit_data
 
-        log.warning("No previous visit found with non-empty values for %s", fields)
+        log.warning("Initial visit has non-empty values for %s", ignore_empty_fields)
```
hewanadun: If we want these fields to be non-empty in the IVP packet, this warning needs to be modified as "No initial visit found with non-empty values for %s"

hewanadun (changes requested):

PR #280, opened 2025-08-22
Update/revised authorizations

bjkeller:
This is a draft of changes to handling of user authorizations for the purposes of review.

Changes are:
* User authorization report from REDCap is now loaded using Pydantic model validation, and aliases used to convert to field names that can be parsed for study and data type.
* Simplify authorization model to have a list of activity names. Also, change name to make more clear that the authorizations are for a study.
* Change user entry to contain a list of study authorization models


hewanadun (commented):
@bjkeller a couple of questions
- If a user has access to ADRC program and one more affiliated studies, aren't they represented by different groups in Flywheel with different ADCIDs? I'm not clear how this is handled with only one ADCID in NACC Directory (and UserEntry)
- Is enrollment project supposed to be always under ADRC study (even if the user does not have access to P30)?
- Do we still need to collect auth email in directory if it is not used in registry now? Although it is defined as optional attribute in `UserEntry`, there are a couple of places in the code that checks whether auth_email is present.

gear/user_management/src/python/user_app/run.py
```diff
@@ -189,8 +189,11 @@ def __get_user_queue(self, user_file_path: str) -> UserQueue[UserEntry]:
         user_list: UserQueue[UserEntry] = UserQueue()
         for user_doc in object_list:
```
hewanadun, line 190: Add a check to skip entries with `permissions_approval = False`

bjkeller: those are already filtered out at this point, but I did add another check after the object is loaded.

PR #285, opened 2025-08-27
Map ADCID to pipeline

bjkeller:
In the legacy system, an ADCID had to be assigned to create a separate submission pipeline.
So, we have a small number of ADCIDs that are not centers, but designators for pipelines.
Making a group for each ADCID has implications all the way through the system to the directory that we don't want to perpetuate.

So, these changes allow an ADCID to be set for a particular ingest pipeline, which allows for the legacy "pipeline" ADCIDs to be assigned to pipelines within a center group with a different ADCID.

> The use of `adcid` is not changed, and refers to a center.
> The use of `pipeline_adcid` refers to the ADCID of a pipeline pipeline.


hewanadun (dismissed):
Looks good!

PR #287, opened 2025-08-28
Attribute Curator: Propagate NP, Optimize

echeng06:
Corresponds to 1.4.2 of `nacc-attribute-deriver`: https://github.com/naccdata/nacc-attribute-deriver/pull/64

Main intent of this change is to back-propagate NP/UDS derived variables to their corresponding forms. However, this is actually built on the work I had done for 1.5.0 of the attribute deriver (which includes many, many more derived variables) in order to include various optimizations done to the gear, so apologies that this PR is larger than just that. 1.4.2 of the attribute deriver itself is more of a "hot-fix" to tie up loose ends on MQT.

CHANGELOG:
* Updates `nacc-attribute-deriver` to `1.4.2`, which exposes curation rules by scope
* Updates backpropagation logic to handle NP and UDS-specific derived variables
* Adds CLS, FTLD, and LBD scopes (scope not yet used in `1.4.2` of `nacc-attribute-deriver` and ignored)
* Reports version of the attribute deriver for better tracking
* Updates to pass `subject_table` (FW subject.info) around heap evaluation instead of reloading on every file, reducing API calls
    * Also ensures/enforces subject metadata is not updated until ALL files have been curated
* Compiles regex for faster execution


common/src/python/curator/regression_curator.py
```diff
@@ -61,7 +61,18 @@ def compare_as_lists(self, value: str, expected: str) -> bool:
             return False
 
         # compare as sets
-        return set(value_as_list) == set(expected_as_list)
+        try:
+            return set(value_as_list) == set(expected_as_list)
+        except TypeError as e:
+            # means dicts - we assume lists of dicts are ordered, so
+            # if it failed the earlier == then it doesn't match
```
bjkeller, line 68: replace "==" with "equality test"

bjkeller (changes requested):
Just a couple of minor details for method names using "process" and a comment

common/src/python/curator/regression_curator.py
echeng06: @bjkeller Sorry I'm not clear what you mean here?

echeng06: Ah I see I thought you were referring to the actual line of code at first - updated!

bjkeller (approved):
Thank you for the changes

PR #288, opened 2025-08-29
Participant Transfers

hewanadun:
Implement participant transfers:
- Update recording transfer requests in identifier-provisioning gear
- Add participant-transfer gear to process recorded transfer requests, request need to be reviewed by a NACC Admin and approved before it can be processed


bjkeller (approved):
This looks good.

PR #293, opened 2025-09-05
Fix/no study center for pipeline

bjkeller:
Places a constraint on creation of ingest pipelines for affiliated studies where enrollment is separate and no pipeline ADCID is provided for the center in the study file.


hewanadun (approved):
Looks good! I noted a few places that docstrings need to be updated.

PR #299, opened 2025-09-18
File Distribution Gear

echeng06:
Sets up gear that distributes (copies) a file to a target project for each ADCID.


bjkeller (approved):
I think this is OK to merge.

Though, I'd like to look for ways to avoid adding code to the utils package

PR #368, opened 2026-02-11
File Distribution: Associate CSV with regex

echeng06:
This feels like a weird/hack solution, so open to other ideas.

An issue with the original distribution gear is that it distributes to every center no matter what. Ideally we only distribute to centers for which the file is applicable. To figure out which are applicable, we have to associate the distributed file with another one, usually a CSV run through `csv-center-splitter`, to pull an associated ADCID list.

The `associated_csv_regex` basically defines a capture regex that runs on the input filename, and searches for another CSV file in the project that matches the capture. If found, the gear will then attempt to read that CSV file for ADCIDs.

This mainly requires the other file is uploaded before/at the same time as the current input file.


bjkeller (commented):
I'm not sure it is any better, but this is a tag-based approach. I think in scenarios we face it is generally OK to assume that we will never be distributing more than one pair of files at a time.  But, this only works for a pair of files like the biomarkers subject/control data. If you had more files to distribute it would require being more clever.

[tag-based-distribution-proposal.md](https://github.com/user-attachments/files/25267810/tag-based-distribution-proposal.md)

PR #392, opened 2026-04-14
Add delete pipeline to form-scheduler

hewanadun:
Add deletion pipeline to process delete requests
Capture and log successful delete event
Update to process finalization and deletion queues by subject instead of one file at a time
Add deletion email notifications


common/src/python/keys/keys.py
```diff
@@ -53,6 +53,7 @@ class DefaultValues:
     UDS_T_PACKET = "T"
     SUBMISSION_PIPELINE = "submission"
```
bjkeller, line 54: I don't think this is a change that needs to be made immediately, but we are using the word "pipeline" differently. I use it to mean the carrier for the datatype, so "form pipeline" or "imaging pipeline". The word I would use here would be "workflow"; e.g., "submission workflow".
We should probably cleanup the distinction at some point

gear/form_scheduler/src/python/form_scheduler_app/form_scheduler_queue.py
```diff
@@ -581,75 +679,265 @@ def _process_pipeline_queue(
...
+            gear_input_info: gear input configuration by locator type
+            gear_inputs: pre-populated fixed/module gear inputs (not mutated)
+        """
+        for file in batch:
+            log.info(
+                "Triggering %s for %s",
+                pipeline.starting_gear.gear_name,
+                file.name,
+            )
+            file_gear_inputs = dict(gear_inputs)
```
bjkeller, line 842: this is making a shallow copy, so please make sure nothing gets modified that shouldn't

gear/form_scheduler/src/python/form_scheduler_app/form_scheduler_queue.py
```diff
@@ -581,75 +679,265 @@ def _process_pipeline_queue(
...
+            event_capture_callback: optional callback to capture events
+            subject_extractor: callable that extracts a subject ID from a
+                filename (differs between finalization and deletion)
+        """
+        # Group files by subject, preserving per-subject sort order
+        subject_queues: Dict[str, List[FileEntry]] = defaultdict(list)
+        for file in subqueue:
+            subject = subject_extractor(file.name, pipeline.name)
+            if not subject:
+                log.warning(
```
bjkeller, line 951: does this need to do anything to prevent the file from being revisited? or is it a problem if it is?

gear/form_scheduler/src/python/form_scheduler_app/form_scheduler_queue.py
```diff
@@ -37,13 +37,34 @@
...
 
+def get_subject_from_input_file(filename: str, pipeline: PipelineType) -> Optional[str]:
+    """Extract subject ID from the input filename."""
+    try:
+        if pipeline == DefaultValues.FINALIZATION_PIPELINE:
+            # filename format: <naccid>_<....>_<module>.json
+            return filename.split("_")[0]
+        elif pipeline == DefaultValues.DELETION_PIPELINE:
+            # filename format: delete_<ptid>_<....>_<module>.json
+            return filename.split("_")[1]
```
bjkeller: would this work if the PTID has an underscore?

gear/form_scheduler/src/python/form_scheduler_app/event_accumulator.py
```diff
@@ -177,23 +185,17 @@ def _create_visit_event(
...
             datatype=pipeline_label.datatype,
-            # module=visit_metadata.module,
-            # packet=visit_metadata.packet,
             data_identification=visit_metadata,
-            timestamp=qc_completion_time,
+            timestamp=completion_time,
         )
 
-    def capture_events(self, json_file: FileEntry, project: ProjectAdaptor) -> None:
+    def capture_qc_event(self, json_file: FileEntry, project: ProjectAdaptor) -> None:
```
bjkeller, line 199: not necessarily asking for a change now, but I think we should move the success checks and visit metadata extraction out of this class. the EventAccumulator shouldn't be responsible for knowing about QC vs Delete events

bjkeller (changes requested):
I think the only potential issue is the underscore splitting in the filename to find the PTID, but please look at the other comments.

Also, docs/form_scheduler/index.md needs some updates. Please add details on 
* The three pipeline types (deletion, submission, finalization) and their processing order
* Module ordering significance (UDS first for submission/finalization, last for deletion)
* The sequential vs. subject-parallel processing distinction

gear/form_scheduler/src/python/form_scheduler_app/form_scheduler_queue.py
hewanadun: Thanks for catching this, updated to support underscore in PTIDs

gear/form_scheduler/src/python/form_scheduler_app/form_scheduler_queue.py
hewanadun: updated to remove the tags

gear/form_scheduler/src/python/form_scheduler_app/form_scheduler_queue.py
hewanadun: shallow copy is fine here, just replacing the gear input files

gear/form_scheduler/src/python/form_scheduler_app/event_accumulator.py
hewanadun: If we remove those functionality, there is nothing much left in the EventAccumulator class, so I'm not clear of the purpose of the EventAccumulator then. We could create subclasses of EventAccumulator to handle QC and Delete events?

bjkeller: probably. let's not worry about it. we will probably come back to this with other pipelines

bjkeller (approved):
good to merge

PR #394, opened 2026-04-16
Fix pulling visit metadata in submission pipeline gears

hewanadun:
Fix typo in FW proxy delete_file method
Fix errors in creating DataIdentification objects in submission pipeline gears
Add module column to enrollment CSV file
Increment gear versions to deploy error log naming format updates


gear/form_qc_checker/src/python/form_qc_app/enrollment.py
```diff
@@ -167,7 +171,7 @@ def visit_row(self, row: Dict[str, Any], line_num: int) -> bool:
                 record=row, line_number=line_num
```
bjkeller, line 173: is there a reason to not use the enriched row here?

bjkeller: also in the write

bjkeller (dismissed):
The only question I have is about whether the enrollment row with the inserted module should be used in other places. Or, I guess whether it is a problem to use it in other places.

gear/form_qc_checker/src/python/form_qc_app/enrollment.py
hewanadun: Didn't add it as module is not part of enrollment form error checks/rule definition. But I think we can set it to optional in the rule definition and add it to the row.

bjkeller:
I think there are some bugs here. tests illustrating them are in this branch https://github.com/naccdata/flywheel-gear-extensions/tree/fix/enrollment-csv-extra-fields

PR #411, opened 2026-05-01
Add pipeline-event-logger gear and refactor image-identifier-lookup

bjkeller:
## Summary

This branch adds the new `pipeline-event-logger` gear and makes supporting refactors to shared event capture infrastructure and `image-identifier-lookup`.

### New: pipeline-event-logger gear

A generic gear that bridges upstream pipeline gears (which write QC results to file metadata) and NACC's project-level QC status logging and S3 event capture systems. It:

- Reads QC outcomes from `file.info.qc.{upstream_gear_name}` using a generic `GearQC`/`GearQCResult` model (no form-specific assumptions)
- Extracts errors via config-driven field mapping (`error_configs`) — supports arbitrary upstream gear error formats (DICOM QC, JSON schema validation, form validation, etc.)
- Updates the project-level QC status log, attributed to the upstream gear
- Optionally captures a `VisitEvent` to S3 based on configurable outcome-to-action mapping
- Supports `dry_run` mode for testing pipeline configuration

Includes full test suite (unit + property-based tests) and documentation.

### Refactor: extract create_visit_event as reusable function

Extracted `create_visit_event()` from `form_scheduler`'s `event_accumulator.py` into `common/src/python/event_capture/event_generator.py` so both `pipeline-event-logger` and `form_scheduler` can use it without duplication.

### Refactor: image-identifier-lookup data_identification write

- Simplified `ImageIdentifierLookup.run()` return type (removed tuple, errors handled as side-effect)
- Updated `run.py` to write `file.info.data_identification` metadata so downstream gears (like `pipeline-event-logger`) can consume it
- Bumped to version 0.2.0

### Config-driven error extraction

Replaced the form-specific `FileQCModel` error reading in `pipeline-event-logger` with a generic approach:
- `ErrorFieldMapping` and `QCErrorConfig` Pydantic models describe how to map gear-specific error fields to `FileError` objects
- Field values are resolved as key lookups against the source dict; if no matching key exists, treated as literals
- Supports multiple check results per gear (e.g., `dicom-validator` + `jsonschema-validation`)

## Files changed

- **39 files**, +4408 / -250 lines
- New gear: `gear/pipeline_event_logger/` (source, tests, docker, docs)
- Shared: `common/src/python/event_capture/event_generator.py`
- Modified: `gear/image_identifier_lookup/`, `gear/form_scheduler/`
- Specs: `.kiro/specs/pipeline-event-logger/`, `.kiro/specs/image-identifier-data-identification-write/`

## Testing

- All existing tests pass
- New test suite for pipeline-event-logger (unit + property-based)
- image-identifier-lookup tests updated for refactored interface
- Fix, lint, and type check all pass


gear/form_scheduler/src/python/form_scheduler_app/event_accumulator.py
```diff
@@ -235,11 +183,12 @@ def capture_qc_event(self, json_file: FileEntry, project: ProjectAdaptor) -> Non
             timestamp = (
                 json_file.info.get("validated-timestamp") if json_file.info else None
             )
-            visit_event = self._create_visit_event(
+            visit_event = create_visit_event(
                 action=ACTION_PASS_QC,
                 visit_metadata=visit_metadata,
                 project=project,
                 completion_time=timestamp if timestamp else qc_log_file.modified,
+                gear_name="form-scheduler",
```
hewanadun: If the gear_name supposed to be the name of the gear that produced the event, use "form-qc-checker" here? Else, update the gear_name in capture_delete_event also to "form-scheduler".

common/src/python/event_capture/event_generator.py
```diff
@@ -28,18 +94,8 @@ def __init__(self, project: ProjectAdaptor) -> None:
...
-            # PipelineLabel uses a wrap validator that accepts strings
-            self._pipeline_label = PipelineLabel.model_validate(self._project.label)
-        except (ValidationError, ValueError) as error:
-            log.warning(
-                "Failed to parse project label %s: %s", self._project.label, error
-            )
-
         # Get pipeline ADCID
         try:
             self._pipeline_adcid = self._project.get_pipeline_adcid()
```
hewanadun, line 102: It seems the cached pipeline_adcid is not used when creating a visit event?

common/src/python/event_capture/event_generator.py
```diff
@@ -13,13 +15,77 @@
...
+        return None
+
+
 class EventGenerator:
-    """Creates VisitEvent objects from extracted log data."""
+    """Creates VisitEvent objects from extracted log data.
+
+    Caches parsed project metadata (pipeline label and ADCID) to avoid
+    re-parsing when creating multiple events for the same project.
+    """
```
hewanadun, line 88: pipeline label is not cached now

gear/pipeline_event_logger/src/docker/manifest.json
```diff
@@ -0,0 +1,65 @@
...
+    "input_file": {
+      "description": "The file processed by the upstream gear. Must have file.info.qc.{upstream_gear_name} and file.info.data_identification populated.",
+      "base": "file"
+    }
+  },
+  "config": {
+    "upstream_gear_name": {
+      "description": "Name of the upstream gear whose QC results to read",
+      "type": "string"
+    },
```
hewanadun, line 38: Need to add "error_configs" to manifest

hewanadun (changes requested):
Two concerns,
- Possibility of concurrent updates to QC status log file
    - Is there only one image file for a given <PTID, visit date, modality>? If there are multiple files, it will result in conflicts when creating/updating the project level log file.
    - pipeline-event-logger can be concurrently triggered by multiple upstream gears on the same input file, which can result in concurrent updates to the same project level log file.
- AFAIK "object" type didn't work in manifest and we had to use JSON string earlier (check csv_subject_splitter gear). Not sure whether "object" is supported now, check with FW.

Looks good otherwise, I added some minor comments inline.

bjkeller:
the pipeline this will be applied to only has one file but it goes through multiple steps, each of which would require logging. Added the job poll to sequence executions using job number to resolve ties.

also changed the manifest to take the "object" as a json string

hewanadun (approved):
Looks good!

PR #431, opened 2026-05-29
Add formver_split option to gather_form_data gear

brendanrsmith:
When formver_split=true, ModuleDataGatherer buckets rows by form version (using the formver field in the merged form data) and emits one CSV per (module, formver) pair instead of a single CSV per module.

Each bucket maintains its own StringCSVWriter, so column sets are naturally restricted to that formver's columns.

Default is false (backward compatible). Existing participant_list consumers and project-mode invocations without the flag are unaffected.

- Add formver_label() helper for filename-safe version labels
- Add split_by_formver kwarg + content_by_formver property to ModuleDataGatherer; raises AttributeError on mode mismatch
- Plumb through GatherFormDataVisitor and ProjectModeVisitor
- _write_module_output branches on gatherer.split_by_formver:
    False: {study_id}-{module}-{date}.csv (existing)
    True:  {study_id}-{module}-{formver_label}-{date}.csv
- Tests for new behavior; defensive updates to existing mocks to tolerate the new constructor kwarg

### DIFF
--- a/common/src/python/data_requests/data_request.py
+++ b/common/src/python/data_requests/data_request.py
@@ -39,28 +39,108 @@ class DataRequestMatch(BaseModel):
     project_label: str
 
 
+def formver_label(formver: Any) -> str:
+    """Normalize a form version value into a filename-safe label.
+
+    Examples:
+        "1"   -> "v1"
+        "1.0" -> "v1"
+        "1.5" -> "v1.5"
+        "3.0" -> "v3"
+        ""    -> "unknown"
+        None  -> "unknown"
+
+    Args:
+      formver: the raw form version value (any type; coerced via str)
+    Returns:
+      a label suitable for use in a filename, e.g. "v3" or "unknown"
+    """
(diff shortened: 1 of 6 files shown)


hewanadun:
@brendanrsmith @bjkeller Few comments
- Instead of maintaining a separate writer and dict of writers to handle form versions, can simplify to use only a dict of writers (i.e. `self.__writers: dict[str, StringCSVWriter]`) with a writer under "default" label for the default path.
- In `ProjectModeConfig::must_have_valid_modules` method, valid modules list is restricted to UDS, LBD, FTLD, is this intentional?
- In `ProjectModeVisitor.run()` method should raise `GearExecutionError` if group or project cannot be found?

brendanrsmith:
@hewanadun addressing these points: 
- unified the dict with a default label per your suggestions
- limited allowlist was carried over from another template gear, I've removed it. If we want to have a limited allowlist we should probably pull that out to a shared util
- yes, added `GearExecutionError` for missing project or group.

PR #438, opened 2026-06-16
Initial REDCap Image Form Creator Gear

davidohunt:
initial REDCap image form creator for sandbox testing


bjkeller (commented):
## PR Review

### Fixed in latest commit (afeac18b)

- GearEngine classmethod call convention
- Missing type annotations on all functions
- Typo and f-string formatting issues
- Trailing newlines on Dockerfile/manifest.json
- Missing test BUILD file
- FileOrigin stub missing `__getitem__`

### Items to address before merge

**1. Commented-out code in run.py**

Several blocks of commented-out code (`edc_project_pid`, session_id from job config). Should either be removed or converted to TODOs with context explaining why they're deferred and when they'll be needed.

**2. `inspect_acquisitions` — `file` used after loop (potential bug)**

Line 164: `if file.modality == "MR"` references the loop variable after the `for file in acq.files` loop exits. If the last acquisition has zero files, this raises `UnboundLocalError`. Even when it works, it only checks the modality of the last file in the last acquisition — is that the intended semantics?

**3. `######### BEGIN/END SECTION FOR COMMON` block**

~130 lines explicitly marked as belonging in a shared module. If this logic will be reused (e.g., by other imaging gears), it should move to `common/` before this pattern proliferates. If it's genuinely one-off, remove the markers.

### Design observations (non-blocking)

**4. `tag_fail` always raises but tags only in non-dry-run**

This means dry_run fails fast on the first validation issue rather than collecting all problems. If the goal of dry_run is to preview what *would* happen, consider a mode that accumulates errors and reports them all.

**5. Race condition in `generate_unique_record_id`**

The sleep/retry approach for uniqueness is a known limitation. Worth documenting in code comments or the gear's README so future maintainers understand the tradeoff.

bjkeller:
@davidohunt I made some changes, and commented on more complex things above. There is a lint error that popped up, but will leave that for you to resolve

davidohunt:
Thanks for reviewing!  I've been able to debug and refactor a bit today and I think 1-3 should be addressed, so I'll reconcile the differences and continue a bit of clean up before resubmitting.  I know my first time with a gear made for a bit of a mess, but I appreciate you slogging through to get it on track.

bjkeller:
You are working at the boundary of things we've used before, so you might have to add some type stubs to get mypy checks to pass

davidohunt:
Ah, okay -- I'll look into type stubs for the next commit.  Thanks!

PR #439, opened 2026-06-16
PHI coordinator gear

zstarkuw:
Adds the **phi-coordinator** gear and a shared **reader_tasks** common module.

- **`gear/phi_coordinator`** — a scheduled gear that gets imaging reader-task form 
  responses (created by image-pii-detector) and assigns the appropriate tags based on whether the reviewer finds PHI 
  in the image. Tags the task with `phi-coordinator` so future runs skip it. A completed task with no usable answer is reset to `Todo` so it will show as incomplete for the reader.
  
- **`common/src/python/reader_tasks`** — pydantic models + `ReaderTaskClient`
  wrapping the reader-task / protocol / form-response REST endpoints the Flywheel
  SDK does not expose via `FWClient`.


bjkeller (approved):
This is good to go. I made some minor changes.

One is that we need to add `_test` to the end of test directories to avoid package name conflicts in some situations. It didn't affect this gear, but best practice now is to do it in all cases.

PR #440, opened 2026-06-17
Support for optional forms with a release date

hewanadun:
Adds support for optional forms that have a release date, using the new UDS D1c form as the first case. For optional forms with a release date, their submission-status check is skipped for visits predating release, and their columns are dropped from records that predate release and were not submitted.

**Changes:**

**Config model** — common/src/python/configs/ingest_configs.py
New optional ModuleConfigs.release_dates: Optional[FormReleaseConfigs].
FormReleaseConfigs => Dict[str, Dict[str, str]] (packet → form → date) with get_release_date(packet, form).([Example configs file](https://github.com/naccdata/flywheel-gear-extensions/blob/feature/d1c-transformations/gear/form_transformer/data/form-data-module-configs.json))

**Preprocessing** — common/src/python/preprocess/preprocessor.py
`_check_optional_forms_status` extracts the form name from tuple entries and skips the MODE-variable requirement when the visit date precedes the form's release date (read via the module's date_field).

**QC checker** — gear/form_qc_checker/src/python/form_qc_app/definitions.py
`get_optional_forms_submission_status` handles tuple entries when building submission status.
If optional form mode variable not present in the input data, rule definition for that form is not included in validation schema (it is assumed that mode variable presence is validated in the pre-processing checks).

**Transformer** — common/src/python/transform/transformer.py
Refactored FieldFilter into an abstract base with two subclasses:
`VersionMapFilter` — existing version-map drop behavior (renamed).
`ReleaseDateFilter` — new: when visitdate < release_date and mode is not in retain_modes, drops the data fields, header_fields, and the mode field. Honors nofill (rejects the record with an EXCLUDED_FIELDS error if data fields are non-empty; header fields are exempt). 
FieldTransformations now holds List[Union[VersionMapFilter, ReleaseDateFilter]]; pydantic smart-union keeps existing JSON entries valid with no format change. [Example schema](https://github.com/naccdata/flywheel-gear-extensions/blob/feature/d1c-transformations/gear/form_transformer/data/transformation-schemas.json)


gear/form_transformer/data/form-data-module-configs.json
```diff
@@ -89,7 +91,8 @@
                         "b3",
                         "b5",
                         "b6",
-                        "b7"
+                        "b7",
+                        ["d1c", "2026-07-14"]
```
echeng06: I think we should avoid mixing types, especially when they involve nested structures.

It may be better to separate the handling of the release date into a different config field altogether (like form_release_date or something) and check that when pulling the optional forms. It also opens up the ability to handle logic related to a release date for non-optional forms, or provide configs for more complicated logic, if that ever ends up being needed.

gear/form_transformer/data/transformation-schemas.json
```diff
@@ -13,6 +13,19 @@
...
+        {
+            "release_date": "2026-07-14",
+            "mode_field": "moded1c",
+            "nofill": true,
+            "retain_modes": ["1"],
+            "header_fields": [
+                "frmdated1c",
+                "initialsd1c",
+                "langd1c",
+                "d1cnot"
```
echeng06, line 24: Again, not a fan of mixing different configs in the same list. But for this one I'm not really sure of a clean way to go about it otherwise since its a fair bit more complicated.

echeng06 (changes requested):
See comments. Code changes themselves look fine otherwise!

echeng06 (dismissed):
Looks good! Thanks for making those changes.

PR #452, opened 2026-06-30
Adds the **phi-image-removal** gear, the step that follows phi-coordinator in the PHI pipeline.

zstarkuw:
* **`gear/phi_image_removal`** — a file-triggered gear that removes an image confirmed to contain PHI (tagged `PHI-Confirmed` by phi-coordinator), leaving a JSON **tombstone** in its place. For the input file it captures the file's details into a tombstone record, uploads it (the original base name with a `.json` extension) into the same acquisition, tags it `PHI-Tombstone`, and deletes the original image. The actions are ordered so the image is never deleted without a tagged tombstone already in place.


bjkeller (approved):
Looks good.

PR #459, opened 2026-07-10
Refactor form transformations

hewanadun:
**Refactor form transformation schema into separated, categorized types**
The transformation schema mixed two conceptually different transforms in a single flat list per module, relying on Pydantic's fragile "smart-mode" union guessing to tell them apart. This PR restructures both the JSON schema and the processing code so transform types are cleanly separated by category, explicitly tagged, and given consistent domain terminology.

**Changes**
Each module now maps to { field_transformations: [...], form_transformations: [...] } — distinct category lists reflecting scope (field-level filtering vs. whole-form removal).
Every entry carries a transform_type tag ("version_map" / "release_date"), enabling new types per category via a discriminated union later.

Renamed *Filter → *Transformation; added a Transformation base plus FieldTransformation / FormTransformation category ABCs.
Replaced the Dict[module, List[union]] root with TransformationSchema → ModuleTransformations container.
TransformerFactory.create() and Transformation.apply() now take a single ModuleConfigs context object instead of separate date_field / release_dates args; each subclass pulls what it needs. FilterTransformer retained as the pipeline adapter.

Unit + gear tests updated for the new class names, grouped schema, and ModuleConfigs context (reusing the uds_ingest_configs() mock).

### DIFF
--- a/common/src/python/transform/transformer.py
+++ b/common/src/python/transform/transformer.py
@@ -2,9 +2,9 @@
 
 import logging
 from abc import ABC, abstractmethod
-from typing import Any, Dict, List, Optional, Set, Union
+from typing import Any, Dict, List, Literal, Optional, Set
 
-from configs.ingest_configs import FormReleaseDates
+from configs.ingest_configs import ModuleConfigs
 from keys.keys import SysErrorCodes
 from nacc_common.data_identification import (
     DataIdentification,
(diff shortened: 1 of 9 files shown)


echeng06 (dismissed):
Looks good!

bjkeller:
@hewanadun I made some minor changes that overrode Emily's approval. I went ahead and merged

PR #460, opened 2026-07-13
CSV Subject Splitter: Normalize Dates

echeng06:
Add configuration to normalize date fields.


bjkeller (dismissed):
If you are good with the way preserve_case is being handled, this is ready to go

echeng06:
> If you are good with the way preserve_case is being handled, this is ready to go

@bjkeller Updated `preserve_case` to also apply to those header configs! Tested on sandbox and looks good, so merged in.

PR #485, opened 2026-08-12
Capture duplicate-submit events in form-transformer

bjkeller:
## Summary

When form-transformer detects a resubmission via `FormPreprocessor.is_existing_visit()`, it now emits a `"duplicate-submit"` VisitEvent to S3. This gives the reporting layer explicit signal to distinguish intentionally-dropped duplicates from legitimately unfinalized submissions.

The feature is fully opt-in via two new manifest config fields (`event_bucket`, `event_environment`). Existing deployments with no config values behave identically to before.

## Changes

- **common/event_capture/visit_events.py** — Add `"duplicate-submit"` to `VisitEventType` literal and `ACTION_DUPLICATE_SUBMIT` constant
- **gear/form_transformer/src/python/form_csv_app/run.py** — Wire `VisitEventCapture` from config; extract `initialize_event_capture()` as a testable helper
- **gear/form_transformer/src/python/form_csv_app/main.py** — Add `__capture_duplicate_event()` method; capture events in the `is_existing_visit` branch
- **gear/form_transformer/src/docker/manifest.json** — Add optional `event_bucket` and `event_environment` config fields
- **gear/form_transformer/test/python/form_csv_app_test/** — Tests in properly-namespaced directory with conftest fixtures
- **common/test/python/event_capture/** — Unit tests for model acceptance of new action type
- **docs/triage-duplicate-submit-events.md** — Triage document explaining the orphaned submit event problem

## Design decisions

- Event capture placed at the exact detection boundary (one event per duplicate, never re-fired during batch reprocessing)
- Three-layer failure isolation: DataIdentification construction, event capture call, and S3 access each wrapped independently
- Config initialization fails fast on S3 errors (GearExecutionError) but fails open on partial config (warning + disabled)

## Testing

- Config state handling tests exercise the real `initialize_event_capture()` function
- Behavioral tests cover: invalid dates, empty dates, missing config, capture failure isolation, metadata copy re-processing, valid capture with field verification
- All checks pass: lint ✓, mypy ✓, 1801 tests ✓

## Coordination needed

The reporting-lambdas repo needs a corresponding change to add `"duplicate-submit"` to the lambda's `VisitEventType` literal and update the S3 retriever pattern regex. Without that, event files accumulate in S3 but are not consumed during checkpoint accumulation.

### DIFF
--- a/gear/form_transformer/src/python/form_csv_app/main.py
+++ b/gear/form_transformer/src/python/form_csv_app/main.py
@@ -2,6 +2,7 @@
 
 import logging
 from collections import defaultdict
+from datetime import datetime
 from typing import Any, DefaultDict, Dict, List, MutableMapping, Optional, TextIO
 
 from configs.ingest_configs import ModuleConfigs
@@ -12,6 +13,8 @@ from error_logging.qc_status_log_creator import (
     FileVisitAnnotator,
     QCStatusLogManager,
 )
+from event_capture.event_capture import VisitEventCapture
+from event_capture.visit_events import ACTION_DUPLICATE_SUBMIT, VisitEvent
 from flywheel.models.file_entry import FileEntry
 from flywheel.rest import ApiException
 from flywheel_adaptor.flywheel_proxy import ProjectAdaptor
@@ -19,6 +22,8 @@ from inputs.csv_reader import CSVVisitor, read_csv
 from keys.keys import PreprocessingChecks, SysErrorCodes
 from nacc_common.data_identification import (
     DataIdentification,
+    EmptyFieldError,
+    InvalidDateError,
 )
 from nacc_common.error_models import FileQCModel, QCStatus
 from nacc_common.field_names import FieldNames
@@ -53,6 +58,10 @@ class CSVTransformVisitor(CSVVisitor):
         module_configs: ModuleConfigs,
         gear_name: str,
         project: Optional[ProjectAdaptor] = None,
+        event_capture: Optional[VisitEventCapture] = None,
+        center_label: str = "",
+        project_label: str = "",
+        timestamp: Optional[datetime] = None,
     ) -> None:
         self.__module = module
         self.__id_column = id_column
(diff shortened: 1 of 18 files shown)


gear/form_transformer/src/python/form_csv_app/run.py
```diff
@@ -166,6 +225,10 @@ def run(self, context: GearContext) -> None:
                 error_writer=error_writer,
                 gear_name=gear_name,
                 downstream_gears=downstream_gears,
+                event_capture=event_capture,
+                center_label=prj_adaptor.group,
```
hewanadun: prj_adaptor is already passed to this method as destination parameter (might be better to rename?), can use that to extract center_label and project_label instead of adding two extra parameters.

gear/form_transformer/src/python/form_csv_app/main.py
```diff
@@ -181,6 +194,9 @@ def visit_row(self, row: Dict[str, Any], line_num: int) -> bool:
             in self.__module_configs.preprocess_checks
             and self.__preprocessor.is_existing_visit(input_record=transformed_row)
         ):
+            # Capture duplicate-submit event BEFORE adding to existing_visits
+            self.__capture_duplicate_event(transformed_row)
```
hewanadun: Move event capture to `update_existing_visits_error_log` method. In that method, if it fails to find QC metadata in the existing JSON file, the visit is added to the current batch and re-processed (and a qc-pass event will be generated if it passes QC). Visit is skipped from validation only if it has QC metadata.

hewanadun (changes requested):
Need to change where `duplicate-submit` is captured. Please see the comment below.

gear/form_transformer/src/python/form_csv_app/run.py
bjkeller: ugh. should have caught that

gear/form_transformer/src/python/form_csv_app/main.py
bjkeller: done

